PR #7, opened 2023-11-22
Use sciline

jl-wynen:
This is a first draft of a Sciline pipeline. Can you take a look if this is a proper usage of Sciline? I'm going to add tests and more documentation later.

I tested it using
```python
import scipp as sc
import sciline
import ess.diffraction
from ess.diffraction.external import powgen
from ess.diffraction import powder
import scippneutron as scn
from ess.diffraction.grouping import group_by_two_theta, merge_all_pixels

from ess.diffraction.types import *

params = {
    CalibrationFilename: 'PG3_FERNS_d4832_2011_08_24.zip',
    DspacingBins: sc.linspace('dspacing', 0.0, 2.3434, 200, unit='angstrom'),
    Filename[SampleRun]: 'PG3_4844_event.zip',
    Filename[VanadiumRun]: 'PG3_4866_event.zip',
    TwoThetaBins: sc.linspace(dim='two_theta', unit='deg', start=25.0, stop=90.0, num=16),
    ValidTofRange: sc.array(dims=['tof'], values=[0.0, 16666.67], unit='us'),
}
providers = [
    *ess.diffraction.providers,
    *powder.providers,
    *powgen.providers,
]
# providers.remove(merge_all_pixels)
# providers = (*providers, group_by_two_theta)
pl = sciline.Pipeline(
    providers,
    params=params,
)
result = pl.compute(DspacingHistogram)
```
And compared to https://scipp.github.io/ess/instruments/external/powgen/powgen_reduction.html#Vanadium-correction
The results don't quite agree. This is because the old workflow did not crop the vanadium data in tof but the new pipeline does. Without this change, the results are identical.


SimonHeybrock:
Looks good. Can you show the resulting graph?

> The results don't quite agree. This is because the old workflow did not crop the vanadium data in tof but the new pipeline does. Without this change, the results are identical.

Why was the cropping added? Is this a new requirement, or was it an oversight in the old workflow?

jl-wynen:
> Why was the cropping added? Is this a new requirement, or was it an oversight in the old workflow?

I think it was an oversight in the old workflow. I see no reason why the vanadium run would be able to access a higher tof range. In Mantid, the tof bounds are obtained from an isntrument characterization file. Since we don't have those, I hard coded the bounds in the workflow. But then forget to apply them to vanadium.

jl-wynen:
graph

SimonHeybrock:
👍

jl-wynen:
Ready now. Fixes #3

src/ess/diffraction/correction.py
```diff
@@ -62,8 +73,10 @@ def normalize_by_monitor(
 
 
 def normalize_by_vanadium(
-    data: sc.DataArray, *, vanadium: sc.DataArray, edges: sc.Variable
-) -> sc.DataArray:
+    data: MergedPixels[SampleRun],
+    vanadium: MergedPixels[VanadiumRun],
```
SimonHeybrock, line 77: I think in powder, the reduction (pixel merging) is referred to as "focussing", I don't know if that name should be adopted here?

src/ess/diffraction/filtering.py
```diff
@@ -67,3 +69,58 @@ def remove_bad_pulses(
...
+        time-of-flight.
+
+    Returns
+    -------
+    :
+        Cropped data.
+    """
+    return TofCroppedData[RunType](
+        data.bin(tof=tof_range.to(unit=data.coords['tof'].unit))
+    )
```
SimonHeybrock, line 95: Note https://scipp.github.io/user-guide/binned-data/filtering.html#Extract-events-falling-into-a-parameter-interval, which would avoid the need for tof binning, i.e., adding `tof` as a dim.

src/ess/diffraction/filtering.py
```diff
@@ -67,3 +69,58 @@ def remove_bad_pulses(
...
+    ----------
+    data:
+        Input events to be filtered.
+
+    Returns
+    -------
+    :
+        `data` with bad events removed.
+    """
+    # TODO this needs to filter by proton charge once we know how
```
SimonHeybrock, line 118: Is this a new or old todo?

src/ess/diffraction/grouping.py
```diff
@@ -22,4 +34,48 @@ def group_by_two_theta(data: sc.DataArray, *, edges: sc.Variable) -> sc.DataArra
...
+    -------
+    :
+        The input without a `'spectrum'` dimension.
+    """
+    return MergedPixels(data.bins.concat('spectrum'))
+
+
+def finalize_histogram(
+    data: NormalizedByVanadium, edges: DspacingBins
+) -> DspacingHistogram:
```
SimonHeybrock, line 60: This looks like we are normalizing in event mode, which is bad for error handling. Should histogram before vanadium division?

src/ess/diffraction/types.py
```diff
@@ -0,0 +1,118 @@
...
+
+class Filename(sciline.Scope[RunType, str], str):
+    """Filename of a run."""
+
+
+OutFilename = NewType('OutFilename', str)
+"""Filename of the output."""
+
+TwoThetaBins = NewType('TwoThetaBins', sc.Variable)
+"""Bin edges for 2theta."""
```
SimonHeybrock, line 45: I suppose this is for grouping by scattering angle, instead of into a single spectrum? Can you document this?

src/ess/diffraction/types.py
```diff
@@ -0,0 +1,118 @@
...
+    """Total proton charge."""
+
+
+CalibrationData = NewType('CalibrationData', sc.Dataset)
+"""Detector calibration data."""
+
+# This is Mantid-specific and can probably be removed when the POWGEN
+# workflow is removed.
+DetectorInfo = NewType('DetectorInfo', sc.Dataset)
+"""Mapping between detector numbers and spectra."""
```
SimonHeybrock, line 63: Should it be moved into `external/powgen` then?

src/ess/diffraction/types.py
```diff
@@ -0,0 +1,118 @@
...
+
+RawCalibrationData = NewType('RawCalibrationData', sc.Dataset)
+"""Calibration data as loaded from file, needs preprocessing before using."""
+
+
+class RawData(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
+    """Raw data."""
+
+
+class RawDataWithvariances(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
```
SimonHeybrock, line 106:
```suggestion
class RawDataWithVariances(sciline.Scope[RunType, sc.DataArray], sc.DataArray):
```

src/ess/diffraction/uncertainty.py
```diff
@@ -0,0 +1,18 @@
...
+    res = data.copy(deep=False)
+    if res.bins is not None:
+        res.bins.constituents['data'].variances = None
+    else:
+        res.variances = None
+    return RawData[VanadiumRun](res)
+
+
+providers = (drop_variances,)
+"""Sciline providers for handling statistical uncertainties."""
```
SimonHeybrock, line 18: Why are the vanadium errors dropped?

docs/examples/POWGEN_data_reduction.ipynb
```diff
@@ -0,0 +1,424 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "35412429-c039-4591-b8b8-49b11a128b7f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "providers = [\n",
+    "    *ess.diffraction.providers,\n",
+    "    *powder.providers,\n",
```
SimonHeybrock, line 86: This looks a bit confusing, `powder` is a submodule of `ess.diffraction`? What is the split? Do you want to explain to the reader?

docs/examples/POWGEN_data_reduction.ipynb
```diff
@@ -0,0 +1,424 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "6335d026-d526-4ad4-bcda-7270e955d2c1",
+   "metadata": {},
+   "source": [
+    "### Save reduced data to file\n",
+    "\n",
+    "We ultimately need to write the reduced data to a file.\n",
+    "This could be done with the `result` we computed above.\n",
+    "But then we would have to provide all metadata manually.\n",
```
SimonHeybrock, line 174: I suppose there is no metadata here? This is a bit confusing maybe?

docs/examples/POWGEN_data_reduction.ipynb
```diff
@@ -0,0 +1,424 @@
...
+   "execution_count": null,
+   "id": "d24c8bf8-ad59-4211-ae6a-3ee29b0556a3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from ess.diffraction.grouping import group_by_two_theta, merge_all_pixels\n",
+    "\n",
+    "grouping_providers = list(providers)\n",
+    "grouping_providers.remove(merge_all_pixels)\n",
+    "grouping_providers = (*grouping_providers, group_by_two_theta)"
```
SimonHeybrock, line 316: Would it make more sense to have `TwoThetaBins` as an optional input to a single provider, which falls back to merging all pixels of no bins provided? This would avoid the need to modify the provider list.

src/ess/diffraction/correction.py
jl-wynen: Probably

src/ess/diffraction/filtering.py
jl-wynen: It wasn't written down before. I added this provider in order to define a `FilteredData` type and make the tof-cropping explicit in the graph. Alternatively, I could drop this provider and make `crop_tof` return `FilteredData`.

src/ess/diffraction/grouping.py
jl-wynen: Maybe. I didn't want to alter the workflow.

docs/examples/POWGEN_data_reduction.ipynb
jl-wynen: I thought the goal was to also have single crystal code in this project. So I kept the old module structure from `ess`.

docs/examples/POWGEN_data_reduction.ipynb
jl-wynen: The filename, sort of. But yes, there isn't any real metadata. I can change it now and revert later when we write a cif file instead.

src/ess/diffraction/grouping.py
SimonHeybrock: Can you open an issue at least?

docs/examples/POWGEN_data_reduction.ipynb
SimonHeybrock: Yes, swapping a provider does not fix the problem, so a different type would work.

docs/examples/POWGEN_data_reduction.ipynb
jl-wynen: I don't know. I thought that was the reason you suggested a common package.

docs/examples/POWGEN_data_reduction.ipynb
jl-wynen: According to my discussion with Andrew about CIF, it seems that we will have to focus in some way that still needs to be determined. It sounds like grouping all pixels is quite useless. But using 2theta is likely also not a good approach, at least for DREAM.

So I wouldn't put too much thought into finding a good solution now. Since this is not the last step in the pipeline, I'd rather keep the unified output type until we know more.

src/ess/diffraction/filtering.py
jl-wynen: This doesn't work. It seems to do nothing:

Screenshot_20231205_160814

src/ess/diffraction/grouping.py
jl-wynen: Done

src/ess/diffraction/uncertainty.py
jl-wynen: Because we normalise in event mode and that requires broadcasting a vanadium histogram to sample events.

docs/examples/POWGEN_data_reduction.ipynb
SimonHeybrock: My point is: You have `ess.diffraction.providers`, implying there is something common?

src/ess/diffraction/uncertainty.py
SimonHeybrock: Maybe we should not do that?

docs/examples/POWGEN_data_reduction.ipynb
jl-wynen: Yes. Many providers are regular functions that were already defined in the common diffraction part of `ess`. I don't know whether they can be used by SXD. Esp. as providers because the domain types are probably all different.

So what do you suggest? Move everything into `powder`? Or remove `powder` and move everything into `diffraction`?

SimonHeybrock: I think you can leave it, but maybe add a comment?

jl-wynen: Opened an issue to track this and make us re-examine when we add sxd code.

PR #10, opened 2023-12-07
Add `dream.load_nexus`

SimonHeybrock:
@celinedurniak this replaces https://github.com/scipp/ess/pull/199. I have adapted it for the latest file I have (which has sorted detector numbers).

### DIFF
--- a/src/ess/dream/io.py
+++ b/src/ess/dream/io.py
@@ -10,6 +10,7 @@ import scippnexus as snx
 
 def load_nexus(
     path: Union[str, os.PathLike],
+    *,
     load_pixel_shape: bool = False,
     entry: str = 'entry',
     fold_detectors: bool = True,
@@ -24,10 +25,10 @@ def load_nexus(
 
     - Mounting-unit, cassette, and counter roughly correspond to the azimuthal angle
       in the mantle detector. However, counter is reversed in the current files. This
-      may also be the case in the other detectors
+      may also be the case in the other detectors.
     - The endcap detectors have a irregular structure that cannot be fully folded.
-      This is not a problem but not again that the counter may be reversed. It is
-      current not clear if this is a bug in the file.
+      This is not a problem but note again that the counter may be reversed. It is
+      currently not clear if this is a bug in the file.
     - The high-resolution detector has a very odd numbering scheme. The SANS detector
       is using the same, but is not populated in the current files. The scheme
       attempts to follows some sort of physical ordering in space (x,y,z), but it
--- a/src/ess/dream/io.py
+++ b/src/ess/dream/io.py
@@ -59,8 +59,19 @@ def load_nexus(
     with snx.File(path, definitions=definitions) as f:
         dg = f[entry][()]
     dg = snx.compute_positions(dg)
-    if not fold_detectors:
-        return dg
+    return fold_nexus_detectors(dg) if fold_detectors else dg
+
+
+def fold_nexus_detectors(dg: sc.DataGroup) -> sc.DataGroup:
+    """
+    Fold the detector data in a DREAM NeXus file.
+
+    The detector banks in the returned data group will have a multi-dimensional shape,
+    following the logical structure as far as possible. Note that the full structure
+    cannot be folded, as some dimensions are irregular.
+    """
+    dg = dg.copy()
+    dg['instrument'] = dg['instrument'].copy()
     instrument = dg['instrument']
     mantle = instrument['mantle_detector']
     mantle['mantle_event_data'] = mantle['mantle_event_data'].fold(
--- a/src/ess/dream/io.py
+++ b/src/ess/dream/io.py
@@ -1,0 +1,121 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+
+import os
+from typing import Any, Dict, Union
+
+import scipp as sc
+import scippnexus as snx
+
+
+def load_nexus(
+    path: Union[str, os.PathLike],
+    load_pixel_shape: bool = False,
+    entry: str = 'entry',
+    fold_detectors: bool = True,
+) -> sc.DataGroup:
+    """
+    Load an unprocessed DREAM NeXus file.
+
+    See https://confluence.esss.lu.se/pages/viewpage.action?pageId=462000005
+    and the ICD DREAM interface specification for details.
+
+    Notes (2023-12-06):
+
+    - Mounting-unit, cassette, and counter roughly correspond to the azimuthal angle
+      in the mantle detector. However, counter is reversed in the current files. This
+      may also be the case in the other detectors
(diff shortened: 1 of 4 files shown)


tests/dream/io_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.fixture
+def filename():
+    return dream.data.get_path('DREAM_nexus_sorted-2023-12-07.nxs')
+
+
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/monitor_bunker")
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/monitor_cave")
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/polarizer/rate")
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/sans_detector")
```
jl-wynen, line 16: Are you planning to remove those filters when we have a complete file?

SimonHeybrock:
> > (which has sorted detector numbers).
> 
> I didn't follow the discussion. Can we rely on them being sorted from now on?

We might, but we are not: Scipp will dynamically chose a faster or slower grouping algorithms, depending on whether they are sorted or not. There is no risk for correctness due to assumptions.

tests/dream/io_test.py
```diff
@@ -0,0 +1,77 @@
...
+    assert 'instrument' in dg
+    instr = dg['instrument']
+    for name in (
+        'mantle',
+        'endcap_backward',
+        'endcap_forward',
+        'high_resolution',
+        'sans',
+    ):
+        assert f'{name}_detector' in instr
```
jl-wynen, line 28: In the GEANT4 loader, those are called `name`, i.e., without `"_detector"`. We should use the same names in both and I'm in favour of the shorter names in the GEANT4 loader. What do you think?

tests/dream/io_test.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/monitor_cave")
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/polarizer/rate")
+@pytest.mark.filterwarnings("ignore:Failed to load /entry/instrument/sans_detector")
+def test_load_nexus_with_pixel_shape(filename):
+    dg = dream.load_nexus(filename, load_pixel_shape=True)
+    assert 'instrument' in dg
+    instr = dg['instrument']
+    # sans_detector is not populated in the current files
+    for name in ('mantle', 'endcap_backward', 'endcap_forward', 'high_resolution'):
+        assert f'{name}_detector' in instr
```
jl-wynen, line 75:
```suggestion
```
Already tested in line below

src/ess/dream/io.py
```diff
@@ -0,0 +1,122 @@
...
+    *,
+    load_pixel_shape: bool = False,
+    entry: str = 'entry',
+    fold_detectors: bool = True,
+) -> sc.DataGroup:
+    """
+    Load an unprocessed DREAM NeXus file.
+
+    See https://confluence.esss.lu.se/pages/viewpage.action?pageId=462000005
+    and the ICD DREAM interface specification for details.
```
jl-wynen, line 22: This isn't great. Are there publicly accessible resources about this? Can we make these public?

src/ess/dream/io.py
```diff
@@ -0,0 +1,122 @@
...
+    - Mounting-unit, cassette, and counter roughly correspond to the azimuthal angle
+      in the mantle detector. However, counter is reversed in the current files. This
+      may also be the case in the other detectors.
+    - The endcap detectors have a irregular structure that cannot be fully folded.
+      This is not a problem but note again that the counter may be reversed. It is
+      currently not clear if this is a bug in the file.
+    - The high-resolution detector has a very odd numbering scheme. The SANS detector
+      is using the same, but is not populated in the current files. The scheme
+      attempts to follows some sort of physical ordering in space (x,y,z), but it
+      looks partially messed up.
```
jl-wynen, line 35: Should these be issues to help us follow up on them?

src/ess/dream/io.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Parameters
+    ----------
+    path:
+        Path to the NeXus file.
+    load_pixel_shape:
+        If True, load the pixel shape from the file's NXoff_geometry group. This is
+        often unused by would slow down file loading. Default is False.
```
jl-wynen, line 43:
```suggestion
        often unused and would slow down file loading. Default is False.
```

src/ess/dream/io.py
```diff
@@ -0,0 +1,122 @@
...
+    instrument = dg['instrument']
+    mantle = instrument['mantle_detector']
+    mantle['mantle_event_data'] = mantle['mantle_event_data'].fold(
+        dim='detector_number',
+        sizes={
+            'wire': 32,
+            'mounting_unit': 5,
+            'cassette': 6,
+            'counter': 2,
+            'strip': 256,
```
jl-wynen, line 73: These names correspond to the ICD, not GEANT4 names on the confluence page. Are these the ones we want to go with? Should we rename the dims in the GEANT4 loader?

src/ess/dream/io.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    definitions = snx.base_definitions()
+    if not load_pixel_shape:
+        definitions["NXdetector"] = FilteredDetector
+
+    with snx.File(path, definitions=definitions) as f:
+        dg = f[entry][()]
+    dg = snx.compute_positions(dg)
+    if not fold_detectors:
+        return dg
```
jl-wynen, line 63: For readability, can you turn the everything below this line into a separate function and call it here?

tests/dream/io_test.py
jl-wynen: I wanted to know if these are permanent and a problem with the loader or just temporary until we get better files.

src/ess/dream/io.py
SimonHeybrock: Why public? This is an ESS repo, so I think everyone who needs access can get to the info?

src/ess/dream/io.py
SimonHeybrock: I am not sure. It depends on how we will handle versioning in the loader. Will we have to keep it working with the current files, if ECDC makes fixes/changes in the future?

src/ess/dream/io.py
jl-wynen: I don't know how that will work for visitors. But fine.

src/ess/dream/io.py
jl-wynen: Maybe not the current ones as they should be replaced by newer ones, I guess. But ultimately, we will have to keep supporting old ones.

src/ess/dream/io.py
jl-wynen: ok

PR #20, opened 2024-02-14
Add issue template

SimonHeybrock:
Needs refinement, but will merge to see if it works.


jl-wynen:
Looks good. But now the button for plain issues is a bit hidden:
Screenshot_20240214_092056

Should we add a plain template?

SimonHeybrock:
Sure, go ahead! Note that I got feedback on the template and updated it, so we are good to move it to a copier template (see task on board).

PR #39, opened 2024-03-04
Move ess.diffraction into ess.powder

### LINKED ISSUE #9: Package structure for subtechniques
The current intention is that ESSdiffraction contains code for both powder and single crystal. But at this point, no SXD code has been added. Some existing code is provided in the root `ess.diffraction` package while some other functions are in `ess.diffraction.powder`.

It is unclear at this point what code can be shared between the techniques. The current split is based on a guess made in the old `ess` project and likely needs to be revised.

This is not a huge deal for regular functions and classes as those can easily be imported from wherever they are. But for Sciline providers, the split is more important because domain types used in the common `ess.diffraction.providers` must match pipelines in both powder and SXD. It is possible that we need to remove the shared provider list and only provide `ess.diffraction.{powder,sxd}.providers`.

See also this thread: https://github.com/scipp/essdiffraction/pull/7#discussion_r1415310288

jl-wynen:
Fixes #9

I decided to move everything because I don't know what parts can be used by SXD. We can move stuff back into ess.diffraction once we start working on that.


docs/api-reference/index.md
```diff
@@ -1,44 +1,32 @@
 # API Reference
 
-## Module Attributes
+## Esspowder
```
SimonHeybrock, line 3:
```suggestion
## ESSpowder
```

SimonHeybrock (approved):

PR #43, opened 2024-03-11
Fix links

jl-wynen:


src/ess/powder/types.py
```diff
@@ -57,6 +57,11 @@ class Filename(sciline.Scope[RunType, str], str):
 class AccumulatedProtonCharge(sciline.Scope[RunType, sc.Variable], sc.Variable):
     """Total proton charge."""
 
+    # Override the docstring of super().__init__ because if contains a broken link
+    # when used by Sphinx in ESSdiffraction.
```
nvaytet, line 61: What is the broken link you are refering to? I don't follow how this works :-(

jl-wynen: There is a link in `Variable.__init__` to the creation functions: https://scipp.github.io/generated/classes/scipp.Variable.html#scipp.Variable.__init__
This link only works in the scipp docs folder. But when rendering `AccumulatedProtonCharge`, Sphinx shows all attributes of its parent class. That includes the link which doesn't work in the ESSdiffraction docs.

This problem will appear everywhere that we subclass Variable. See also https://github.com/scipp/sciline/issues/143

nvaytet (approved):

PR #47, opened 2024-04-02
Add sans in loader

celinedurniak:
The csv loader for DREAM was changed by adding support for the SANS detector.

New test data can be found at https://project.esss.dk/nextcloud/index.php/s/i2iD2M2EHD4rRnM
It corresponds to the following case:
- sample: Si powder
- High flux configuration of the instrument
- Detectors: mantle with 7 modules, endcap backward and forward with 13 modules and complete layouts for High-resolution and SANS detectors

PR #48, opened 2024-04-03
Exclude detectors w/o data in load_geant4_csv

jl-wynen:
The current implementation includes all detectors in the output even if they are not present in the file. This is because of how the data is grouped.

Found in #47.

### DIFF
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -148,10 +148,10 @@ def _split_detectors(
 def _extract_detector(
     detector_groups: sc.DataArray, detector_id_name: str, detector_id: sc.Variable
 ) -> Optional[sc.DataArray]:
-    try:
-        return detector_groups[detector_id_name, detector_id].value
-    except IndexError:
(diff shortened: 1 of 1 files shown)


src/ess/dream/io/geant4.py
```diff
@@ -148,10 +148,10 @@ def _split_detectors(
 def _extract_detector(
     detector_groups: sc.DataArray, detector_id_name: str, detector_id: sc.Variable
 ) -> Optional[sc.DataArray]:
-    try:
-        return detector_groups[detector_id_name, detector_id].value
-    except IndexError:
+    events = detector_groups[detector_id_name, detector_id].value
+    if len(events) == 0:
         return None
```
SimonHeybrock, line 153: Won't downstream code break if some detectors are missing/none?

jl-wynen: Isn't this is how the nexus loader behaves?
Stuff may break but it seems cleaner to not include a detector in the group if the file also didn't include it.

SimonHeybrock (approved):

Issue #50, opened 2024-04-16
[Requirement] Masking tools


celinedurniak:
### Executive summary

Provide masks for different physical quantities: wavelength, 2 theta, intensity and detectors

### Context and background knowledge

These masks are first to be used in the Powder Diffraction workflow for DREAM
Different physical parameters could be used in order to mask:
- Faulty detectors
- Contribution from SE or instrument component in the detected signal
- Debye-Scherrer rings
- Too-high, too low or uneven distribution of intensity 

**Follow-up:** 
- adapt implementation with DREAM NeXus file
- use widget to determine masks (optional and for some physical quantities only)

### Inputs

For masks in 2 theta, wavelength: DataArray containing the quantity to be masked as coordinate

For masks of detectors: metadata about detectors should be stored (positions, other characteristics like sumo, segment, sector…)

### Methodology

For the detectors, the masks could be specified either by the spatial coordinates of the voxels to mask or by sub-elements (i.e., sumo, sector, wire, strip). It could be a set like bank (hr, sans, endcap fwd, endcap bwd, mantle), sector (for SANS and HR) or sumo (for endcaps), module, segment, wire(s) and or strip(s), counter

### Outputs

Data array with the same structure as input and with added mask.

### Which interfaces are required?

Integrated into reduction workflow, Python module / function

### Test cases

- Mask in two theta, intensity: data from D16 @ ILL [link1](https://project.esss.dk/nextcloud/index.php/apps/files/?dir=/&fileid=39562) or DREAM csv files below
- Detector mask, wavelength: DREAM csv data  (any csv files in [link2](https://project.esss.dk/nextcloud/index.php/apps/files/?dir=/DREAM_simple_pwd_workflow&fileid=17822376))

### Comments

Note: The file form D16 can be used in the online documentation

nvaytet:
So masking should be part of a workflow.
Would you expect the masks to be different for each run/sample during an experiment or can you use a mask that was saved to a file?
I'm trying to figure out how this would fit in a pipeline?

Would you have a set of parameters at the top of your notebook like a list of wavelength or two_theta ranges that would defined what to mask, or do you need to have an interactive tool where you can draw masks?
I guess maybe a mixture of the 2?

The former could be something like (API to be discussed):
```Py
params[TwothetaMasks] = [sc.array(dims=['two_theta'], values=[53, 55], units='deg),
                         sc.array(dims=['two_theta'], values=[80, 84], units='deg)]

[...]

PR #55, opened 2024-05-21
Add masking

nvaytet:
This adds the framework for masking according to various coordinates (`pixel_id`, `tof`, `wavelength`, `two_theta`).

It also attempts to standardize the Powgen and dream-geant4 use cases, in terms of loading files and keeping logical dimensions.

This is part of #50, but it only implements the reduction workflow side of things, and not an interactive masking tool, which is defered to a later PR.

(Apologies about the noise with quotation marks, I accidentally ran `black` at some point on the files, after the config for `black` was removed due to the switch to `ruff`, and it therefore did not ignore quotation marks)


docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -23,40 +23,63 @@
...
     "    # The upper bounds mode is not yet implemented.\n",
     "    UncertaintyBroadcastMode: UncertaintyBroadcastMode.drop,\n",
+    "    # Edges for binning in d-spacing\n",
+    "    DspacingBins: sc.linspace(\"dspacing\", 0.0, 2.3434, 201, unit=\"angstrom\"),\n",
+    "    # Mask in time-of-flight\n",
+    "    TofMask: lambda x: (x < sc.scalar(0.0, unit=\"ns\"))\n",
+    "    | (x > sc.scalar(86e6, unit=\"ns\")),\n",
+    "    # Mask in wavelength\n",
+    "    WavelengthMask: lambda x: (x > sc.scalar(2.0, unit=\"angstrom\"))\n",
+    "    & (x < sc.scalar(2.7, unit=\"angstrom\")),\n",
```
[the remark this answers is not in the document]
nvaytet: No there is no artifact. I was just to illustrate things.
I probably should make a separate example, either in a new section at the bottom or in a new notebook.

nvaytet:
@jl-wynen for now, the notebooks only work when computing d-spacing from positions, and not from a calibration file.
I looked at computing the other scattering coordinates from the d-spacing, if calibration is present, but it seemed more difficult than I thought.
Maybe you know of a good approach for this that I missed?

jl-wynen:
I know an approach but it may not be good. For CIF files we need to write data in tof. So the idea was to do the reduction in d-spacing and then convert back to tof using fake positions. You could use a similar approach here where you essentially determine calibrated positions by converting back from d-spacing. 

There is some ambiguity here about how you define positions and tof, though. So you / we would have to make a judgement call. But since we need this anyway for CIF, we can use the same mechanism here.

src/ess/dream/io/nexus.py
```diff
@@ -1,135 +1,156 @@
...
+    RawSource,
+    ReducibleDetectorData,
+    RunType,
+    SamplePosition,
+    SourcePosition,
+)
+from ess.reduce import nexus
+
+DETECTOR_BANK_RESHAPING = {
+    "endcap_backward_detector": lambda x: x.fold(
```
nvaytet, line 32: Make this into sizes, not lambdas

src/ess/powder/external/powgen/beamline.py
```diff
@@ -6,9 +6,16 @@
...
-from ...types import CalibrationData, DetectorDimensions, RawCalibrationData
+from ...types import (
+    CalibrationData,
+    NeXusDetectorDimensions,
+    NeXusDetectorName,
+    RawCalibrationData,
+)
 from .types import DetectorInfo
 
+DETECTOR_BANK_SHAPES = {"powgen_detector": {"bank": 23, "column": 154, "row": 7}}
```
nvaytet, line 17: shapes -> sizes

src/ess/powder/grouping.py
```diff
@@ -2,67 +2,33 @@
...
     TwoThetaBins,
 )
 
 
-def group_by_two_theta(
-    data: DspacingData[RunType], edges: TwoThetaBins
+def focus_data(
+    data: MaskedData[RunType],
+    dspacing_bins: DspacingBins,
+    twotheta_bins: Optional[TwoThetaBins] = None,
```
nvaytet, line 21: Maybe request a new data type to get data binned in theta, as in sans Qxy?

docs/user-guide/sns-instruments/POWGEN_data_reduction.ipynb
```diff
@@ -50,19 +50,21 @@
...
-    "    ValidTofRange: sc.array(dims=['tof'], values=[0.0, 16666.67], unit='us'),\n",
-    "\n",
     "    # Edges for binning in d-spacing\n",
-    "    DspacingBins: sc.linspace('dspacing', 0.0, 2.3434, 200, unit='angstrom'),\n",
+    "    DspacingBins: sc.linspace(\"dspacing\", 0.0, 2.3434, 201, unit=\"angstrom\"),\n",
+    "    # Mask in time-of-flight\n",
+    "    TofMask: lambda x: (x < sc.scalar(0.0, unit=\"us\"))\n",
+    "    | (x > sc.scalar(16666.67, unit=\"us\")),\n",
+    "    # Mask in wavelength\n",
+    "    WavelengthMask: lambda x: (x > sc.scalar(0.18, unit=\"angstrom\"))\n",
```
nvaytet, line 66: remove wavelength mask

src/ess/dream/data.py
```diff
@@ -2,38 +2,38 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 """Data for tests and documentation with DREAM."""
 
-from pathlib import Path
+# from pathlib import Path
```
nvaytet, line 5: commented code

jl-wynen (approved):

PR #63, opened 2024-06-11
Update for next version of Sciline

SimonHeybrock:
The main change is around the new way of handling `Optional`. Pixel masks are now handled using `map`, which will also support multiple masks.

Note that this is incompatible with the released Sciline, i.e., we will need to make a synchronized release.


src/ess/powder/masking.py
```diff
@@ -71,3 +67,31 @@ def apply_masks(
 
 
 providers = (read_pixel_masks, apply_masks)
+
+
+def _merge(*dicts: dict) -> dict:
+    return {key: value for d in dicts for key, value in d.items()}
+
+
+def set_pixel_mask_filenames(
```
jl-wynen, line 76:
```suggestion
def with_pixel_mask_filenames(
```
because it does not modify the input.

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -53,6 +53,8 @@
     "    # Mask in time-of-flight to crop to valid range\n",
     "    TofMask: lambda x: (x < sc.scalar(0.0, unit=\"ns\"))\n",
     "    | (x > sc.scalar(86e6, unit=\"ns\")),\n",
+    "    TwoThetaMask: None,\n",
+    "    WavelengthMask: None,\n",
```
jl-wynen, line 57: Can we have default params so users don't have to specify everything? I thought that was the plan anyway?

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -91,7 +93,8 @@
     "    *powder.providers,\n",
     ")\n",
     "\n",
-    "pipeline = sciline.Pipeline(providers, params=params)"
+    "pipeline = sciline.Pipeline(providers, params=params)\n",
+    "pipeline = powder.set_pixel_mask_filenames(pipeline, [])"
```
jl-wynen, line 97: It is not great that this always needs to be called even when there are no file names. Can this be done with a default parameter?

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
   "graphviz",
   "plopp",
   "pythreejs",
-  "sciline>=23.9.1",
+  "sciline @ git+https://github.com/scipp/sciline@main",
```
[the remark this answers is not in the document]
SimonHeybrock: Sure that is the plan, this is temporary to pass CI.

src/ess/powder/masking.py
SimonHeybrock: @nvaytet Should we change all the naming of similar functions in ESSsans?

docs/user-guide/dream/dream-data-reduction.ipynb
SimonHeybrock: Yes, but as far as I could tell ESSdiffraction is not implementing C.5 yet (see https://github.com/scipp/essreduce/blob/main/docs/user-guide/reduction-workflow-guidelines.md#c5-use-a-fixed-pattern-for-creating-manipulating-and-running-workflows), so this is unrelated to this PR?

docs/user-guide/dream/dream-data-reduction.ipynb
SimonHeybrock: I agree, but it cannot, unfortunately (since the `Optional` this is replacing is a couple of nodes downstream of the filename).

Alternatives:
- Add parameter to the workflow creation, such `DreamWorkflow(pixel_masks=[])`.
- Set a default for `DetectorMasks`, which has the pitfall that later setting `PixelMaskFilename` siltently has no effect (since setting the default for `DetectorMasks` trimmed the branch of providers leading to `PixelMaskFilename`. `set_pixel_mask_filenames` could re-add this branch, but the pitfall remains.

jl-wynen: What about setting `PixelMaskFilename = []` by default?

SimonHeybrock: You mean as a default param in `DreamWorkflow`? That can be done, but mapping again differently afterwards is not possible.

src/ess/powder/masking.py
nvaytet: fine by me

SimonHeybrock: Done

pyproject.toml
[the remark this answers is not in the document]
SimonHeybrock: Updated.

SimonHeybrock:
As agreed elsewhere, I will merge and release. Main comments were addressed, other things (such as masking UX) can be improved later, once we have a better idea.

PR #65, opened 2024-06-19
feat: implement s2g tranform

### LINKED ISSUE #49: [Requirement] Fourier transform to compute PDF
### Executive summary

We need to compute the fourier transform of the structure factor to get the PDF.

This is needed for reducing PDF data but not other powder data.

The structure factor is given by
```math
S(Q) = \frac{I_\mathsf{coh}(Q) - \sum c_i{|f_i(Q)|}^2}{{|\sum c_i f_i(Q)|}^2} + 1
```
where $I_\mathsf{coh}$ is the coherent scattering intensity, $f_i$ the scattering length density, and $c_i = \rho_i/\rho$ with $\rho_i$ the number density of atoms of type $i$. The momentum transfer is $Q = 4\pi \mathsf{sin}(\theta)/\lambda$.

The PDF is then
```math
g(r) - 1 = A \frac{2}{\pi} \int_{Q_\mathsf{min}}^{Q_\mathsf{max}}\, (S(Q) - 1) Q \mathsf{sin}(Qr) dQ
```
with
```math
A = \frac1{2\pi^2 r \rho \sum b^2}
```

### Inputs

1d $S(Q)$.

### Methodology

The integral that defines $g(r)$ corresponds to a Fourier transform. So to implement it, we should use an FFT. We should probably add support in Scipp itself. The code in ESSdiffraction would just be a simple wrapper to call the FFT.

### Outputs

Fourier transformed, 1d $g(r)$

Python module / function

### Test cases

None yet, though there will be simulated data for DREAM.

### Comments

_No response_

jokasimr:
Fixes #49 

Draft because it needs some tests and some manual testing with data (either fake, or simulated or real). I'll ask the IDS.

Maybe we should also integrate it with the workflow pipeline and add an example using this in the workflow.

### DIFF
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -54,11 +54,14 @@ def compute_pdf_from_structure_factor(
     uncertainty_broadcast_mode: UncertaintyBroadcastMode,
         Choose how uncertainties in S(Q) are broadcast to D(r).
         Defaults to ``UncertaintyBroadcastMode.drop``.
+    return_covariances:
+        bool, if True the second output of the function will be a 2D array representing the covariance
+        matrix of the entries in the first output.
 
     Returns
     -------
     :
-        1D DataArray representing :math:`D(r)` with a bin-edge coordinate called :math:`r` that is the provided output grid
+        1D DataArray representing :math:`D(r)` with a bin-edge coordinate called :math:`r` that is the provided output grid, and optionally a 2D DataArray representing the covariances of :math:`D(r)`.
 
     '''  # noqa: E501
     q = s.coords['Q']
@@ -76,6 +79,6 @@ def compute_pdf_from_structure_factor(
     g = sc.DataArray(g.data, coords={'r': r})
     if return_covariances:
         cov_g = _covariance_of_matrix_vector_product(c * v, ioq)
-        cov_g = sc.DataArray(cov_g, coords={'r': r})
+        cov_g = sc.DataArray(cov_g, coords={d: r for d in cov_g.dims})
         return g, cov_g
     return g
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -29,19 +29,19 @@ def compute_pdf_from_structure_factor(
     '''
     Compute a pair distribution function from a structure factor.
 
-    Computes the pair distribution function :math:`D(r)` as defined in
-    `Review: Pair distribution functions from neutron total scattering for the study of local structure in disordered materials <https://www.sciencedirect.com/science/article/pii/S2773183922000374>`_
+    Computes the pair distribution function :math:`G(r)` as defined in
+    `Review: Pair distribution functions from neutron total scattering for the study of local structure in disordered materials <https://www.sciencedirect.com/science/article/pii/S2773183922000374>`_ (note, in the reference, the pair distribution function is denoted :math:`D(r)`, but since :math:`G(r)` is the more common name, that is what is used here).
     from the overall scattering function :math:`S(Q)`.
 
     The inputs to the algorithm are:
 
     * A histogram representing :math:`S(Q)` with :math:`N` bins on a bin-edge grid with :math:`N+1` edges :math:`Q_j` for :math:`j=0\\ldots N`.
-    * The bin-edge grid over :math:`r` the output histogram representing :math:`D(r)` will be computed on.
+    * The bin-edge grid over :math:`r` the output histogram representing :math:`G(r)` will be computed on.
 
     In each output bin, the output is computed as:
 
     .. math::
-        D_{i+\\frac{1}{2}} &= \\frac{2}{\\pi(r_{i+1}-r_i)} \\int_{r_i}^{r_{i+1}} \\int_{0}^\\infty (S(Q) - 1) Q \\sin(Q r) dQ \\ dr  \\\\
+        G_{i+\\frac{1}{2}} &= \\frac{2}{\\pi(r_{i+1}-r_i)} \\int_{r_i}^{r_{i+1}} \\int_{0}^\\infty (S(Q) - 1) Q \\sin(Q r) dQ \\ dr  \\\\
         &\\approx \\frac{2}{\\pi(r_{i+1}-r_i)} \\sum_{j=0}^{N-1} (S(Q)_{j+\\frac{1}{2}} - 1) (\\cos(\\bar{Q}_{j+\\frac{1}{2}} r_{i})-\\cos(\\bar{Q}_{j+\\frac{1}{2}} r_{i+1})) \\Delta Q_{j+\\frac{1}{2}}
 
     Note that in the above expression the subscript :math:`_{j+\\frac{1}{2}}` is used to denote
@@ -54,7 +54,7 @@ def compute_pdf_from_structure_factor(
     r:
         1D array, bin-edges of output grid
     uncertainty_broadcast_mode: UncertaintyBroadcastMode,
-        Choose how uncertainties in S(Q) are broadcast to D(r).
+        Choose how uncertainties in S(Q) are broadcast to G(r).
         Defaults to ``UncertaintyBroadcastMode.drop``.
     return_covariances:
         bool, if True the second output of the function will be a 2D array representing the covariance
@@ -63,7 +63,7 @@ def compute_pdf_from_structure_factor(
     Returns
     -------
     :
-        1D DataArray representing :math:`D(r)` with a bin-edge coordinate called :math:`r` that is the provided output grid, and optionally a 2D DataArray representing the covariances of :math:`D(r)`.
+        1D DataArray representing :math:`G(r)` with a bin-edge coordinate called :math:`r` that is the provided output grid, and optionally a 2D DataArray representing the covariances of :math:`G(r)`.
 
     '''  # noqa: E501
     q = s.coords['Q']
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -63,10 +63,10 @@ def compute_pdf_from_structure_factor(
         UncertaintyBroadcastMode.drop,
     ):
         # avoid scipp uncertainty propagation exception when multiplying v and ioq
-        # the variances are correct after summing over the `Q` dim
+        # the variances are correct after summing over q.dim
         ioq = ioq.broadcast(sizes=v.sizes).copy()
     else:
         ioq = broadcast_uncertainties(ioq, prototype=v, mode=uncertainty_broadcast_mode)
     c = 2 / sc.constants.pi / dr
-    g = c * (v * ioq).sum('Q')
+    g = c * (v * ioq).sum(q.dim)
     return sc.DataArray(g.data, coords={'r': r})
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -14,7 +14,9 @@ def _covariance_of_matrix_vector_product(A, v):
         A = A.transpose()
     cov = (sc.sqrt(v) * A).values
     cov = cov @ cov.T
-    return sc.array(dims=[A.dims[0], A.dims[0] + '_2'], values=cov)
+    return sc.array(
+        dims=[A.dims[0], A.dims[0] + '_2'], values=cov, unit=v.unit * A.unit**2
+    )
 
 
 def compute_pdf_from_structure_factor(
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -79,6 +79,8 @@ def compute_pdf_from_structure_factor(
     g = sc.DataArray(g.data, coords={'r': r})
     if return_covariances:
         cov_g = _covariance_of_matrix_vector_product(c * v, ioq)
-        cov_g = sc.DataArray(cov_g, coords={d: r for d in cov_g.dims})
+        cov_g = sc.DataArray(
+            cov_g, coords={d: r.rename_dims({r.dim: d}) for d in cov_g.dims}
+        )
         return g, cov_g
     return g
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -3,15 +3,13 @@
 """Signal transformation algorithms for powder diffraction."""
 
 import scipp as sc
-
-from .uncertainty import UncertaintyBroadcastMode, broadcast_uncertainties
+from ess.reduce.uncertainty import UncertaintyBroadcastMode, broadcast_uncertainties
 
 
 def compute_pdf_from_structure_factor(
     s: sc.DataArray,
-    *,
-    rho: sc.Variable,
     r: sc.Variable,
+    *,
     uncertainty_broadcast_mode=UncertaintyBroadcastMode.drop,
 ) -> sc.DataArray:
     '''
@@ -24,7 +22,6 @@ def compute_pdf_from_structure_factor(
     The inputs to the algorithm are:
 
     * A histogram representing :math:`S(Q)` with :math:`N` bins on a bin-edge grid with :math:`N+1` edges :math:`Q_j` for :math:`j=0\\ldots N`.
-    * The density :math:`\\rho`, the total number of atoms per volume.
     * The bin-edge grid over :math:`r` the output histogram representing :math:`D(r)` will be computed on.
 
     In each output bin, the output is computed as:
@@ -40,8 +37,6 @@ def compute_pdf_from_structure_factor(
     ----------
     s:
         1D DataArray representing :math:`S(Q)` with a bin-edge coordinate called :math:`Q`
-    rho:
-        total number of atoms per unit volume
     r:
         1D array, bin-edges of output grid
     uncertainty_broadcast_mode: UncertaintyBroadcastMode,
@@ -63,7 +58,7 @@ def compute_pdf_from_structure_factor(
     v = v[r.dim, :-1] - v[r.dim, 1:]
 
     ioq = (s - sc.scalar(1.0, unit=s.unit)) * dq
-    ioq = broadcast_uncertainties(ioq, uncertainty_broadcast_mode)
+    ioq = broadcast_uncertainties(ioq, prototype=v, mode=uncertainty_broadcast_mode)
     c = 2 / sc.constants.pi / dr
     g = c * (v * ioq).sum('Q')
     return sc.DataArray(g.data, coords={'r': r})
--- a/src/ess/powder/transform.py
+++ b/src/ess/powder/transform.py
@@ -1,0 +1,69 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+"""Signal transformation algorithms for powder diffraction."""
+
+import scipp as sc
+
+from .uncertainty import UncertaintyBroadcastMode, broadcast_uncertainties
+
+
+def compute_pdf_from_structure_factor(
+    s: sc.DataArray,
+    *,
+    rho: sc.Variable,
+    r: sc.Variable,
+    uncertainty_broadcast_mode=UncertaintyBroadcastMode.drop,
+) -> sc.DataArray:
+    '''
+    Compute a pair distribution function from a structure factor.
+
+    Computes the pair distribution function :math:`D(r)` as defined in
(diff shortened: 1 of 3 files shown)


jokasimr:
### Some very preliminary results

I've applied the transformation to the results in the `dream-data-reduction.ipynb` example notebook file.

The issue instructions says 
```math
S(Q) = \frac{I_\mathsf{coh}(Q) - \sum c_i{|f_i(Q)|}^2}{{|\sum c_i f_i(Q)|}^2} + 1
```
but I've taken $S(Q)$ to be $I(Q)$ instead where $I(Q)$ is the sample intensity divided by the vanadium intensity.

Here's what $I(Q)$ looks like 
intensity
and here are the resulting $g(r)$ when using the transform with/without the sinc-filter:
distribution-with-filter
distribution-without-filter

I don't know what the expected results are or what kind of sample this data comes from (it's simulated right?).

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    * The bin-edge grid over :math:`r` the output histogram representing :math:`D(r)` will be computed on.
+
+    In each output bin, the output is computed as:
+
+    .. math::
+        D_{i+\\frac{1}{2}} = \\frac{2}{\\pi(r_{i+1}-r_i)} \\int_{r_i}^{r_{i+1}} \\int_{0}^\\infty (S(Q) - S_{self}) Q sin(Q r) dQ \\ dr  \\\\
+        \\approx \\frac{2}{\\pi(r_{i+1}-r_i)} \\sum_{j=0}^{N-1} (S(Q)_{j+\\frac{1}{2}} - S_{self}) (cos(\\bar{Q}_{j+\\frac{1}{2}} r_{i})-cos(\\bar{Q}_{j+\\frac{1}{2}} r_{i+1})) \\Delta Q_{j+\\frac{1}{2}}
+
+    Note that in the above expression the subscript :math:`_{j+\\frac{1}{2}}` is used to denote
+    quantities belonging to the :math:`j`:th bin of a histogram, :math:`\\bar{Q}_{j+\\frac{1}{2}} = \\frac{Q_j + Q_{j+1}}{2}` and :math:`\\Delta Q_{j+\\frac{1}{2}} = Q_{j+1} - Q_{j}`.
```
jl-wynen, line 287:
```suggestion
    quantities belonging to the :math:`j^\\text{th}` bin of a histogram, :math:`\\bar{Q}_{j+\\frac{1}{2}} = \\frac{Q_j + Q_{j+1}}{2}` and :math:`\\Delta Q_{j+\\frac{1}{2}} = Q_{j+1} - Q_{j}`.
```
Because `j:th` is Swedish spelling, as far as I can tell.

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+
+    '''  # noqa: E501
+    for i in range(s.size):
+        if not sc.isnan(s[i]).value:
+            minbound = i
+            break
+    for i in range(s.size - 1, -1, -1):
+        if not sc.isnan(s[i]).value:
+            maxbound = i
+            break
```
jl-wynen, line 315: Why is this needed?

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+
+    s = s[minbound : maxbound + 1]
```
jl-wynen, line 317: This functions requires `s` to be 1D. Can this be relaxed easily? If not, please document the requirement!

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    q = s.coords['Q']
```
jl-wynen, line 318: Hard coded name. Please either make it an argument (with a default) or document the coord name.

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    qm = sc.midpoints(q)
```
jl-wynen, line 319: This is only correct for linearly spaced bins. Handling other types of bins is difficult here because you need both bin centres and bin edges. So please add a `sc.islinspace(q)` check and raise an exception. We can implement support for other bins later if needed.

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    dq = q[1:] - q[:-1]
+    dr = r[1:] - r[:-1]
+
+    v = sc.cos(qm * r * sc.scalar(1, unit='rad'))
+    v = v[r.dim, :-1] - v[r.dim, 1:]
+
+    ioq = (4 * sc.constants.pi * rho) * (s - self_scattering)
+    ioq.variances = None
```
jl-wynen, line 327: Why? Looks like you should have a `UncertaintyBroadcastMode` argument to determine the behaviour.

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    s:
+        :math:`S(Q)` with bin-edge coordinate :math:`Q`
+    rho:
+        total number of atoms per unit volume
+    r:
+        bin-edges of output grid
+    self_scattering:
+        the amplitude of the overall scattering that is due to self-scattering
+    use_filter:
+        if ``True`` the Lorch filter is applied
```
jl-wynen, line 300: Please define the filter in maths notation.

tests/powder/correction_test.py
```diff
@@ -315,3 +319,43 @@ def test_apply_lorentz_correction_needs_coords():
...
+def test_pdf_structure_factor_needs_q_coord():
+    da = sc.DataArray(sc.ones(sizes={'Q': 3}))
+    r = sc.array(dims='r', values=[2, 3, 4, 5.0])
+    with pytest.raises(KeyError):
+        compute_pdf_from_structure_factor(
+            da, sc.scalar(1.0, unit='1/angstrom'), r, use_filter=False
+        )
+
+
+@pytest.mark.parametrize('use_filter', [(False,), (True,)])
```
jl-wynen, line 333:
```suggestion
@pytest.mark.parametrize('use_filter', [False, True])
```

tests/powder/correction_test.py
```diff
@@ -315,3 +319,43 @@ def test_apply_lorentz_correction_needs_coords():
...
+        sc.array(dims='Q', values=[1, 2, 4.0], unit='angstrom**4'),
+        coords={'Q': sc.array(dims='Q', values=[0, 1, 2, 3.0], unit='1/angstrom')},
+    )
+    r = sc.array(dims='r', values=[2, 3, 4, 5.0], unit='angstrom')
+    v = compute_pdf_from_structure_factor(
+        da, sc.scalar(1.0, unit='1/angstrom**3'), r, use_filter=True
+    )
+    sc.testing.assert_allclose(
+        v.data,
+        sc.array(dims='r', values=[-4.25706, -0.528801, -5.29488], unit='1/angstrom'),
```
jl-wynen, line 359: Where do those numbers come from?

src/ess/powder/correction.py
```diff
@@ -257,6 +257,86 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    Computes the pair-distribution-function :math:`D(r)` as defined in
+    `Review: Pair distribution functions from neutron total scattering for the study of local structure in disordered materials <https://www.sciencedirect.com/science/article/pii/S2773183922000374>`_
+    from the overall scattering function :math:`S(Q)`.
+
+    The inputs to the algorithm are:
+
+    * A histogram representing :math:`S(Q)` with :math:`N` bins on a bin-edge grid with :math:`N+1` edges :math:`Q_j` for :math:`j=0\\ldots N`.
+    * The density :math:`\\rho`, the total number of atoms per volume.
+    * The structure independent coherent scattering amplitude :math:`S_{self} = \\sum_m c_m b_m^2` where the sum is over the atomic species in the sample, :math:`c_m` is the fraction of atoms that belong to species :math:`m` and :math:`b_m` is the size of scattering from an atom of that species.
+    * The bin-edge grid over :math:`r` the output histogram representing :math:`D(r)` will be computed on.
```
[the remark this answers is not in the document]
jokasimr: Yes it felt a bit like that to me to. But it also seemed reasonable to have the definitions before the actual description of the procedure. Otherwise the user would have to go to the parameter section to understand what the symbols in the expression for $D$ means. It seemed easier to just have some duplication here.

src/ess/powder/correction.py
jokasimr: Good to know, thanks.

src/ess/powder/correction.py
jokasimr: When running the diffraction workflow (on dream data I think that was) there are some values near the edges of the Q-range where the intensities are `nan`. The code above selects a range of `Q` where the intensities are not `nan` before proceeding. Every `S(Q)` value contributes to every `D(r)` value so if we don't extract a `nan` free range then every `D(r)` value is just `nan`.

src/ess/powder/correction.py
jokasimr: Should we support non-1d `s`? I'm not sure if it can be relaxed, that depends on what a n-d `s` means.

src/ess/powder/correction.py
jokasimr: What do you mean by "correct"?

I don't think an exception is the best way to go here. Using the midpoint is not incorrect for non-equidistant q if by correct we mean "as the q-binning is refined, the result converges to the exact integral".
Using the midpoint here is an approximation both when the q edges are linspaced and when they are unequally spaced. It's a *better* approximation (in a sense) when the q edges are linspaced.

Requiring linspaced grid here limits the usability of the routine without any quantifiable correctness improvement.

tests/powder/correction_test.py
jokasimr: I think that didn't work for some reason. I'll try it again.

tests/powder/correction_test.py
jokasimr: They're what the routine returns for the test-data in the test right now.
It doesn't mean anything, it just checks if the result has changed.

jokasimr:
I don't think the Lorch filtering makes sense. Seems like it just convolves the output $D(r)$ with a flat window function. Applying the filter or using `np.convolve(pdf_default.data.values, np.ones(K)/K, 'same')` looks virtually identical if you choose an appropriate `K`. That indicates to me it doesn't actually deal with the spurious oscillations, it just sweeps them under the rug so to speak, and while doing that it also smooths the physical component of `D(r)`.

Maybe that doesn't matter, if the subsequent data-analysis doesn't care about the widths of the peaks in `D(r)` then the smoothing procedure maybe doesn't do much harm.
But unless it's something that is explicitly asked for, so far it isn't, I think we should skip it.

src/ess/powder/correction.py
jl-wynen: There could be separate `S(Q)` for a number of `two_theta` bands. I don't know whether that is needed in PDF data reduction. So for now, we can keep the 1d requirement. I would just like it to be explicit.

src/ess/powder/correction.py
jokasimr: Yes that sounds reasonable, I'll change it

jl-wynen:
> I don't think the Lorch filtering makes sense. Seems like it just convolves the output with a flat window function. Applying the filter or using np.convolve(pdf_default.data.values, np.ones(K)/K, 'same') looks virtually identical if you choose an appropriate K. That indicates it doesn't actually deal with the spurious oscillations, it just sweeps them under the rug so to speak, and while doing that it also smooths the physical component of D(r).

This sounds like what people write elsewhere. They smooth out the curve to remove oscillations.
I guess it depends on whether the change to the physical signal is less bad than having the spurious oscillations.

> But unless it's something that is explicitly asked for, so far it isn't, I think we should drop the Lorch filtering from this PR.

:+1: I agree. Remove it from this PR. We can always add it later if required.

src/ess/powder/transform.py
```diff
@@ -0,0 +1,69 @@
...
+    ----------
+    s:
+        1D DataArray representing :math:`S(Q)` with bin-edge coordinate :math:`Q`
+    rho:
+        total number of atoms per unit volume
+    r:
+        1D array, bin-edges of output grid
+    uncertainty_broadcast_mode: UncertaintyBroadcastMode,
+        Choose how uncertainties in S(Q) are broadcast to D(r).
+        Defaults to ``UncertaintyBroadcastMode.drop``.
```
jl-wynen, line 49: I think `fail` is the better, more defensive default. I'd rather do the safer alternative by default and require intervention to use a riskier option.

src/ess/powder/correction.py
jl-wynen: Ok. Leave it

jl-wynen:
Can you try and figure out the value of $\rho$ that you need to reproduce the example? And then plot the results in the same figure as well as their difference? We should do a proper comparison to check that we are doing the right thing.

src/ess/powder/transform.py
jokasimr: Won't that make the function fail 100% of the time unless the user manually specifies the `uncertainty_broadcast_mode` to be something different? Seems excessively safe and not very useful.

The application of a linear operator on a vector will cause the entries of the output vector to be correlated, the only way we can avoid correlation in this case is by not invoking the function.

If we want the user to be aware that the output data is not uncorrelated maybe it's better to display a warning.

jokasimr:
> Can you try and figure out the value of that you need to reproduce the example? And then plot the results in the same figure as well as their difference? We should do a proper comparison to check that we are doing the right thing.

:+1: I have a meeting with Celine on Thursday, I'll bring that up with her.

src/ess/powder/transform.py
jl-wynen: Warnings are almost always ignored. Only hard errors get people's attention.

Everywhere else, the default is to raise on variance broadcasts. So why is this function different?

> Won't that make the function fail 100% of the time unless the user manually specifies the uncertainty_broadcast_mode to be something different?

Or when users remove variances early. But in practice, people may never want to drop them but use the upper bound method instead.

jokasimr: I think this kind of protecting the users from themselves is a mistake. This makes the user having to specify the `uncertainty_propagation_mode` argument every time this function is called (if the input has variances, which I think is the common case). That is an annoying UX.

We must assume the users know to some extent what they are doing. This function does not support (the classical type of) uncertainty propagation. Of course that does not mean we should return erroneous or misleading results, that's why I'm suggesting making it the default to `drop` variances.

Dropping the variances will not cause erroneous results. Worst case scenario it causes an error later when the result doesn't have variances but the user expected it to have them. I think that is better than the user having to specify the `uncertainty_propagation_mode` argument every time this function is called.

SimonHeybrock: I am inclined to say that if the result has no error bars at all, then making that the default is ok. If the output *does* have error bars, but some *contribution* to them is dropped, doing that by default is problematic.

src/ess/powder/transform.py
```diff
@@ -0,0 +1,72 @@
...
+    qm = sc.midpoints(q)
+    dq = q[1:] - q[:-1]
+    dr = r[1:] - r[:-1]
+
+    v = sc.cos(qm * r * sc.scalar(1, unit='rad'))
+    v = v[r.dim, :-1] - v[r.dim, 1:]
+
+    ioq = (s - sc.scalar(1.0, unit=s.unit)) * dq
+    if uncertainty_broadcast_mode in (
+        UncertaintyBroadcastMode.fail,
```
jl-wynen, line 62: Needs a third case to handle failure.

src/ess/powder/transform.py
```diff
@@ -0,0 +1,72 @@
...
+        UncertaintyBroadcastMode.drop,
+    ):
+        # avoid scipp uncertainty propagation exception when multiplying v and ioq
+        # the variances are correct after summing over q.dim
+        ioq = ioq.broadcast(sizes=v.sizes).copy()
+    else:
+        ioq = broadcast_uncertainties(ioq, prototype=v, mode=uncertainty_broadcast_mode)
+    c = 2 / sc.constants.pi / dr
+    g = c * (v * ioq).sum(q.dim)
```
jl-wynen, line 71: Please compute the covariance matrix. If it is simple enough, return it. If not, still compute, e.g., with ScippUncertainty, and find out how large the correlations are.

jl-wynen (changes requested):

src/ess/powder/transform.py
jokasimr: I'll just call the uncertainty broadcast function like this https://github.com/scipp/essdiffraction/pull/65/files#diff-ddeaa8aa9904e068ac3fc37753a0c2122c59a05047b3734453a84fcb8d03fb1eR74
and let it handle the cases.

That should be fine right?

jokasimr:
Visualization of the covariance matrix:
cov

jl-wynen:
Looks like low correlations. But you can compute a correlation matrix (Pearson r) and show `r - np.eye(r.shape[0])`?

SimonHeybrock:
Is it expected that there are negative values in the correlation matrix?

jokasimr:
> Is it expected that there are negative values in the correlation matrix?

It's expected. A negative covariance basically means, given that this entry increased it's likely the other entry decreased. That could happen if there's, for example, uncertainty in the position of a peak but not it's overall shape. Moving the peak increases some entries and decreases others. The entries affected will have a negative covariance.

jokasimr:
> Looks like low correlations. But you can compute a correlation matrix (Pearson r) and show r - np.eye(r.shape[0])?

Please explain why you want me to do that

jl-wynen:
> Please explain why you want me to do that

Because the correlation matrix is easier to interpret. The covariances matrix can only be interpreted in combination with the mean values. The correlation matrix has values in `[-1, 1]`. And I asked for the `-eye` to remove self-correlations which would make the plot harder to read.

jokasimr:
Here is a visualization of the correlation matrix minus the diagonal:
corr

and the same but zoomed in on the lower left corner:
corr_minus_self_zoomed_in

jl-wynen:
So we have basically 100% correlation between ~10 neighbours for most elements. And fairly strong anti-correlations for slightly farther away elements. This on its own does not mean that follow up operations are wrong if they neglect correlations. But it seems to strongly suggest so.

We should follow up on it with scientists and see if their fits change when they take correlations into account. But for now, I think we should definitely return the covariance matrix.

jl-wynen (approved):

jokasimr:
>  So we have basically 100% correlation between ~10 neighbours for most elements. And fairly strong anti-correlations for slightly farther away elements. This on its own does not mean that follow up operations are wrong if they neglect correlations. But it seems to strongly suggest so.

Strong correlations between neighboring elements is expected.
I'd expect the width (in terms of pixels) of the region of strong correlation should grow with the density of the output grid (the `r`-grid).
That there are strong correlations across ~10 neighbors here just indicates I've used a too dense output grid.

Having a too-dense output grid is of course not good, but it would surprise me if later fitting procedures were badly affected by it, unless they are fitting very high frequency components of `r`, and that seems unlikely.

src/ess/powder/transform.py
```diff
@@ -29,19 +29,19 @@ def compute_pdf_from_structure_factor(
     '''
     Compute a pair distribution function from a structure factor.
 
-    Computes the pair distribution function :math:`D(r)` as defined in
-    `Review: Pair distribution functions from neutron total scattering for the study of local structure in disordered materials <https://www.sciencedirect.com/science/article/pii/S2773183922000374>`_
+    Computes the pair distribution function :math:`G(r)` as defined in
+    `Review: Pair distribution functions from neutron total scattering for the study of local structure in disordered materials <https://www.sciencedirect.com/science/article/pii/S2773183922000374>`_ (note, in the reference, the pair distribution function is denoted :math:`D(r)`, but since :math:`G(r)` is the more common name, that is what is used here).
```
jokasimr, line 33: Celine asked for a name change

jl-wynen: Works for me.

PR #66, opened 2024-06-19
Helper and notebook for removing vanadium peaks

### LINKED ISSUE #24: [Requirement] Strip vanadium peaks
### Executive summary

Removes Vanadium peaks out of a background by linearly/quadratically interpolating over the expected peak positions.

This algorithm attempts to find all Vanadium peaks and subtract them from the data, leaving just the ‘background’.

The data will then be used to normalize "sample" data collected with the same instrumental configuration.

### References

https://docs.mantidproject.org/nightly/algorithms/StripVanadiumPeaks-v2.html?highlight=stripvanadiumpeaks

#### Assumptions

  -  this method will be used for the reduction of powder diffraction data
  -  the input data should correspond to measurements or simulations with a Vanadium sample (information contained in the log of the NeXus file)
  -  this first implementation considers 1D data
  -  there should be a choice for the background shape between linear and quadratic
  -  the input data should be expressed in d-spacing
  -  for the implementation of algorithms for finding and stripping peaks, check NMX reduction workflow. 

#### Preconditions

  -  algorithms to find and strip peaks
  -  input data in d-spacing

### Inputs

  -  Scipp data array of Vanadium

  -  List of diffraction peaks for Vanadium (d-spacing positions in Å). 

### Methodology

The theoretical positions of the peaks are $\frac{a}{\sqrt{h^2+k^2+l^2}}$ with a≈3.03Å and h, k, l integers and h+k+l even 

See Mantid's implementation

### Outputs

Scipp data array of Vanadium containing only background information

Python module / function

### Test cases

- csv file data_dream_vanadium_inc_coh.csv.zip in this [folder](https://project.esss.dk/nextcloud/index.php/apps/files/?dir=/DREAM_simple_pwd_workflow&fileid=17822376) (source_position=[-3.478 mm, 0, -76550 mm], sample_position=[0, 0, 0]) or PG3_733.nxs from the Usage Data of Mantid

-  list of Vanadium peaks (see Input section above)

### Comments

_No response_

jl-wynen:
Fixes #24.

I only added it in the POWGEN module because the default fit parameters depend on the data. And we don't have DREAM data that could be used here. But it should be straight forward to transfer it to DREAM once we have that data.


docs/user-guide/sns-instruments/POWGEN_data_reduction.ipynb
```diff
@@ -111,6 +111,27 @@
...
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "afb476bb-71f9-4ac5-b6aa-0a5e35895f6a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "d = pipeline.compute(FocussedDataDspacing[VanadiumRun])\n",
```
nvaytet, line 121: This could probably do with some explanation as to what it is plotting?

docs/user-guide/sns-instruments/POWGEN_data_reduction.ipynb
```diff
@@ -386,7 +407,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.12"
+   "version": "3.10.13"
   }
  },
  "nbformat": 4,
```
nvaytet, line 413: I'm not sure I see peak fitting or removal used in the notebook? Am I missing something?

docs/user-guide/sns-instruments/vanadium_processing.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "id": "84f8b41c-e3f3-4282-83c6-e0e424bc5632",
+   "metadata": {},
+   "source": [
+    "## Removing coherent scattering peaks\n",
+    "\n",
+    "As the variable name `peaked_data` implies, the produced spectrum contains peaks from coherent scattering.\n",
+    "Even though the peaks are small for vanadium, we need to remove them to extract pure incoherent scattering.\n",
+    "We can approximate the coherent scattering contribution by fitting functions to the peaks and subtracting those fitted functions.\n",
+    "[scippneutron.peaks](https://scipp.github.io/scippneutron/generated/modules/scippneutron.peaks.html) contains general functionality for fitting and removing peaks.\n",
+    "Here, we use it through [ess.powder.external.powgen.peaks](../../generated/modules/ess.powder.external.powgen.peaks.rst) which provides useful defaults for vanadium peaks at POWGEN.\n",
```
nvaytet, line 139: Once #68 will be merged, the path will be `ess.snspowder.powgen`. The files should also be moved around. Hopefully this won't be too annoying.

docs/user-guide/sns-instruments/vanadium_processing.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   "id": "19251c69-75eb-4dd0-be97-d2272946bacd",
+   "metadata": {},
+   "source": [
+    "Remove the fitted peaks to obtain the incoherent scattering.\n",
+    "Also restore the bin-edge coordinate that we had to replace temporarily for the fits.\n",
+    "\n",
+    "Importantly, we remove variances from the data.\n",
+    "If we kept the variances, subtracting the fitted models would introduce correlations between the data points.\n",
+    "This corresponds to [UncertaintyBroadcastMode.drop](../../generated/modules/ess.powder.types.UncertaintyBroadcastMode.rst) in the main workflow.\n",
+    "See also the [guide in ESSreduce](https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#s-8-propagation-of-uncertainties-in-broadcast-operations-should-support-drop-and-upper-bound-strategies-upper-bound-shall-be-the-default)"
```
nvaytet, line 223:
```suggestion
    "See also the [guide in ESSreduce](https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#s-8-propagation-of-uncertainties-in-broadcast-operations-should-support-drop-and-upper-bound-strategies-upper-bound-shall-be-the-default)."
```

docs/user-guide/sns-instruments/vanadium_processing.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fbcde35f-8608-4ecb-9b6d-67f686d7839a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "incoherent = scn.peaks.remove_peaks(sc.values(to_fit), fit_results)\n",
```
nvaytet, line 233: [Unrelated to this PR] Are the `fit_vanadium_peaks` and `scn.peaks.remove_peaks` designed to be used in pairs, on the same data, or could you imagine have some `fit_results` from another run and you use them to `remove_peaks` on a new Vanadium data?
I guess the height of the peaks etc would become an issue?

docs/user-guide/sns-instruments/vanadium_processing.ipynb
```diff
@@ -0,0 +1,395 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6f19a754-c41a-4507-9ac5-a9f68dbab60f",
+   "metadata": {},
+   "source": [
+    "We can further inspect the results.\n",
+    "Below, there is a function that plots\n",
+    "- the data with coherent and incoherent scattering (blue),\n",
```
nvaytet, line 245: New line before list for sphinx rendering

docs/user-guide/sns-instruments/vanadium_processing.ipynb
nvaytet: Yes and it is also useful for making the diagnostics plot.

Should we add those functions into a workflow graph so we have a workflow which computes a "vanadium without peaks" or something like that?

docs/user-guide/sns-instruments/vanadium_processing.ipynb
```diff
@@ -0,0 +1,381 @@
...
+    "    if xlim is not None:\n",
+    "        def in_range(x: sc.Variable) -> bool:\n",
+    "            return sc.isfinite(x) and (xlim[0] <= x) and (x < xlim[1])\n",
+    "        data = data[data.dim, xlim[0]:xlim[1]]\n",
+    "        removed = removed[removed.dim, xlim[0]:xlim[1]]\n",
+    "        fit_results, peak_estimates = zip(*(\n",
+    "            (r, e)\n",
+    "            for r, e in zip(fit_results, peak_estimates, strict=True)\n",
+    "            if in_range(r.window[0]) and in_range(r.window[1])\n",
+    "        ), strict=True)\n",
```
jl-wynen, line 275: I changed this because the behaviour suddenly changed on my machine. I don't know what caused it.
Now, when using `ax.set_xlim` the text shown with `ax.text` is still shown. This messes up how the axes are shown in a figure. The new solution removes everything that is not within the shown range. It may remove a fit that lies at the edge of the range, though.

docs/user-guide/sns-instruments/vanadium_processing.ipynb
nvaytet: In the reflectometry workflow, johannes added some types which were plots instead of data. You could then `pipeline.compute(DiagnosticPlot)` to view the figure.
Maybe something like this coudl also work here?

That said, I'm not so bothered, leaving it as is also works.

jl-wynen: This could be done for diagnostics. But we may still need to modify the fit results before removing peaks. I'd say we leave it as is until we have more insight. In particular from DREAM.

nvaytet (approved):

Issue #60, opened 2024-05-29
Compute positions, wavelength, 2theta from calibration


nvaytet:
In #55 the computation of d-spacing from calibration was effectively removed because we needed the wavelength, 2theta coordinates for adding masks.
We should be able to compute them from the final d-spacing, by first altering the detector positions?

SimonHeybrock:
Not sure what you are suggesting, but I don't think this is possible in general? Firstly, calibration gives a single parameter (d-spacing), so full position cannot be defined, and secondly, I think there is (or at least can be) non-linear terms in the "d-spacing from calibration" definition, so it is not clear you can reverse and compute the raw parameter values.

jl-wynen:
You cannot reverse the transform to get the true positions back. But we can do a liner transformation from d back to tof by setting some artificial time offset and positions. I don't like this a bit. But it is needed for writing CIF files anyway.

jl-wynen:
## Context
We currently compute wavelength and 2theta using the 'normal' way without calibration. We then use those values to add masks and focus into 2theta bins. This makes the resulting 2theta and d-spacing coords inconsistent. To fix this, we would have to compute wavelength and 2theta (and potentially positions) using calibration.

The usual powder calibrations account for a number of factors such as errors in measured detector positions and the time of flight and wavelength profile of the beam. Since all those are combined into the calibration parameters, it is unclear how to disentangle them to obtain calibrated wavelength and positions separately.

## Potential solution: leave it as is
It may be that the differences between results with and without calibration are below the required precision for wavelength and 2theta. In this case, we can keep using the current solution. I think this is what Mantid does.
We need input from the instrument team to decide.

PR #67, opened 2024-06-20
Compute d-spacing from calibration

jl-wynen:
Part of #60

This computes d-spacing from calibration. But wavelength and 2theta are still computed as before. This means that d and lambda,2theta are not consistent. We don't know yet whether that is a problem in practice. It may not be because we only need wavelength and 2theta for masking and coarse binning. So we may be able to get away with less accuracy in those quantities.


jl-wynen:
This was implemented in collaboration with @nvaytet. So I am skipping the review.

PR #68, opened 2024-06-21
Add workflow helper

nvaytet:
Adds `PowgenWorkflow()` and `DreamGeant4Workflow()` helpers.

I also moved the `ess/powder/external` folder to `ess/snspowder` to be more like `ess/isissans` in `esssans`.


src/ess/dream/__init__.py
```diff
@@ -20,8 +33,35 @@
...
+    }
+
+
+def DreamGeant4Workflow() -> sciline.Pipeline:
+    """
+    Workflow with default parameters for the Powgen SNS instrument.
+    """
+    return sciline.Pipeline(
+        providers=powder_providers + geant4_providers, params=default_parameters()
+    )
```
SimonHeybrock, line 58: - Does this belong into the `__init__.py`?
- Docstring refers to Powgen.

docs/api-reference/index.md
SimonHeybrock, line 1: Shouldn't the workflows be listed in the api-reference?

SimonHeybrock (commented):
Shoul

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -17,92 +17,70 @@
...
+    "workflow[CalibrationFilename] = None\n",
+    "workflow[NeXusDetectorName] = \"mantle\"\n",
+    "# The upper bounds mode is not yet implemented.\n",
+    "workflow[UncertaintyBroadcastMode] = UncertaintyBroadcastMode.drop\n",
+    "# Edges for binning in d-spacing\n",
+    "workflow[DspacingBins] = sc.linspace(\"dspacing\", 0.0, 2.3434, 201, unit=\"angstrom\")\n",
+    "# Mask in time-of-flight to crop to valid range\n",
+    "workflow[TofMask] = lambda x: (x < sc.scalar(0.0, unit=\"ns\")) | (x > sc.scalar(86e6, unit=\"ns\"))\n",
+    "workflow[TwoThetaMask] = None\n",
+    "workflow[WavelengthMask] = None\n",
```
jl-wynen, line 72: Can those be `None` by default?

src/ess/dream/__init__.py
nvaytet: > * Does this belong into the `__init__.py`?

hmm in `essans`, we have a separate [workflow file](https://github.com/scipp/esssans/blob/main/src/ess/loki/workflow.py) for loki, in `essreflectometry`, for Amor, we have the [workflow](https://github.com/scipp/essreflectometry/blob/main/src/ess/amor/__init__.py) in the `__init__.py`. We should be consistent and just have a separate workflow file?

> * Docstring refers to Powgen.

Bad copy/paste ;-)

docs/user-guide/dream/dream-data-reduction.ipynb
nvaytet: They could, but I thought this would make it obvious what the syntax is if you want to add other masks. I guess this goes back to the discussion on whether these notebooks should only show what is needed, and we need another page that describes everything that can be done in the workflow.

src/ess/dream/__init__.py
SimonHeybrock: I don't think consistency is the main argument, I thought the `__init__.py` should generally be kept minimal?

SimonHeybrock (approved):

PR #71, opened 2024-06-25
Performance

### LINKED ISSUE #41: Check workflow performance
Need to ensure that we can process the expected data volume with the available memory and fast enough for Beamline.

SimonHeybrock:
### Changes

- Bypass the same single-threading problem encountered in SANS, I think we either have to put a helper into ESSreduce, or see if the underlying problem should be fixed in Scipp.
- Avoid storing position for every event. This will only affect Geant4 data, so the improvements from that will not be seen for NeXus data. Note that this can lead to NaN positions if there are no events. Is this a problem (not in the tests)? If so we either need to revert, or find a better solution.
- Fixes #41 (but I guess it has to be revisited sooner or later). I did not spot any other "obvious" problems in the functions taking the most time.

### Baseline

Baseline for Geant4 workflow:

```
2.47 seconds load_geant4_csv
4.75 seconds load_geant4_csv
3.00 seconds extract_geant4_detector_data
6.11 seconds extract_geant4_detector_data
0.62 seconds normalize_by_proton_charge
4.38 seconds add_scattering_coordinates_from_positions
0.35 seconds apply_masks
0.99 seconds normalize_by_proton_charge
4.36 seconds convert_to_dspacing
5.20 seconds add_scattering_coordinates_from_positions
0.23 seconds apply_masks
2.94 seconds convert_to_dspacing
9.75 seconds focus_data_dspacing
10.12 seconds focus_data_dspacing
1.19 seconds normalize_by_vanadium_dspacing
```

I artificially increased the data size by modifying the data after loading:

```python
def extract_geant4_detector_data(
    detector: RawDetector[RunType],
) -> RawDetectorData[RunType]:
    """Extract the histogram or event data from a loaded GEANT4 detector."""
    data = extract_detector_data(detector)
    data = data.bins.concatenate(data)
    data = data.bins.concatenate(data)
    data = data.bins.concatenate(data)
    data = data.bins.concatenate(data)
    data = data.bins.concatenate(data)
    data = data.bins.concatenate(data) 
    data = data.bins.concatenate(data)  # ~ 10 GByte (sample run)
    return RawDetectorData[RunType](data)
```

The timings reported for this function are thus irrelevant in reality.

### This branch

`RawDetectorData[SampleRun]` is now 6 GByte instead of 10 GByte:
```
2.45 seconds load_geant4_csv
5.28 seconds load_geant4_csv
1.97 seconds extract_geant4_detector_data
4.03 seconds extract_geant4_detector_data
0.34 seconds normalize_by_proton_charge
0.20 seconds convert_to_dspacing
2.14 seconds focus_data_dspacing
0.46 seconds normalize_by_proton_charge
0.23 seconds apply_masks
0.20 seconds convert_to_dspacing
2.14 seconds focus_data_dspacing
0.94 seconds normalize_by_vanadium_dspacing
```

### DIFF
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -167,15 +167,11 @@ def _extract_detector(
     return events
 
 
-def get_source_position(
-    raw_source: RawSource[RunType],
-) -> SourcePosition[RunType]:
+def get_source_position(raw_source: RawSource[RunType]) -> SourcePosition[RunType]:
     return SourcePosition[RunType](raw_source["position"])
 
 
-def get_sample_position(
-    raw_sample: RawSample[RunType],
-) -> SamplePosition[RunType]:
+def get_sample_position(raw_sample: RawSample[RunType]) -> SamplePosition[RunType]:
     return SamplePosition[RunType](raw_sample["position"])
 
 
@@ -184,10 +180,11 @@ def patch_detector_data(
     source_position: SourcePosition[RunType],
     sample_position: SamplePosition[RunType],
 ) -> ReducibleDetectorData[RunType]:
-    out = detector_data.copy(deep=False)
-    out.coords["source_position"] = source_position
-    out.coords["sample_position"] = sample_position
-    return ReducibleDetectorData[RunType](out)
+    return ReducibleDetectorData[RunType](
+        detector_data.assign_coords(
+            source_position=source_position, sample_position=sample_position
+        )
+    )
 
 
 def geant4_detector_dimensions(
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -102,8 +102,10 @@ def _group(detectors: Dict[str, sc.DataArray]) -> Dict[str, sc.DataGroup]:
     def group(key: str, da: sc.DataArray) -> sc.DataArray:
         if key in ["high_resolution", "sans"]:
             # Only the HR and SANS detectors have sectors.
-            return da.group("sector", *elements)
-        res = da.group(*elements)
+            res = da.group("sector", *elements)
+        else:
+            res = da.group(*elements)
+        res.coords['position'] = res.bins.coords.pop('position').bins.mean()
         res.bins.coords.pop("sector", None)
         return res
 
--- a/src/ess/powder/grouping.py
+++ b/src/ess/powder/grouping.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 """Grouping and merging of pixels / voxels."""
 
+import scipp as sc
+
 from .types import (
     DspacingBins,
     DspacingData,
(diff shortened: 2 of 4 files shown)


SimonHeybrock:
Converted to draft, since it looks like I messed up memory consumption (at least CI is having trouble, it seems).

SimonHeybrock:
Looking at https://github.com/scipp/essdiffraction/actions/workflows/ci.yml it seems the CI timings were a perfectly normal outlier. Ready for review once more!

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -181,9 +181,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "intermediates[MaskedData[SampleRun]].bins.concat().hist(\n",
-    "    two_theta=300, wavelength=300\n",
-    ").plot(norm=\"log\")"
+    "two_theta = sc.linspace(\"two_theta\", 0.8, 2.4, 301, unit=\"rad\")\n",
```
jl-wynen, line 184: Why the explicit limits? Does this really matter so much for performance?

src/ess/dream/io/geant4.py
```diff
@@ -102,8 +102,10 @@ def _group(detectors: Dict[str, sc.DataArray]) -> Dict[str, sc.DataGroup]:
     def group(key: str, da: sc.DataArray) -> sc.DataArray:
         if key in ["high_resolution", "sans"]:
             # Only the HR and SANS detectors have sectors.
-            return da.group("sector", *elements)
-        res = da.group(*elements)
+            res = da.group("sector", *elements)
+        else:
+            res = da.group(*elements)
+        res.coords['position'] = res.bins.coords.pop('position').bins.mean()
```
jl-wynen, line 108: Did you check whether this changes the final result?

src/ess/powder/grouping.py
```diff
@@ -12,23 +14,42 @@
 )
 
 
+def _drop_grouping_and_bin(
+    data: sc.DataArray, *, dims_to_reduce: tuple[str, ...] | None = None, edges: dict
+) -> sc.DataArray:
+    all_pixels = data if dims_to_reduce == () else data.bins.concat(dims_to_reduce)
```
jl-wynen, line 20:
```suggestion
def _drop_grouping_and_bin(
    data: sc.DataArray, *, edges: dict
) -> sc.DataArray:
    all_pixels = data.bins.concat(dims_to_reduce)
```
because `dims_to_reduce` is never used.

src/ess/powder/grouping.py
```diff
@@ -12,23 +14,42 @@
...
+    # all_pixels may just have a single bin now, which currently yields
+    # inferior performance when binning (no/bad multi-threading?).
+    # We operate on the content buffer for better multi-threaded performance.
+    if all_pixels.ndim == 0:
+        content = all_pixels.bins.constituents['data']
```
jl-wynen, line 25: Is it safer to use this? Because it excludes constituents that are out of range?
```suggestion
        content = all_pixels.value
```

jl-wynen, line 25: It seems like this should be done by `bin` automatically. Is this possible?

docs/user-guide/dream/dream-data-reduction.ipynb
SimonHeybrock: Because we have some voxels with NaN positions, and get an exception otherwise.

src/ess/dream/io/geant4.py
SimonHeybrock: I was hoping there are tests for that...?

src/ess/powder/grouping.py
SimonHeybrock: I was using it for the two-theta case, but then refactored. Since I wanted to consider moving this to ESSreduce I thought I'd keep it.

src/ess/powder/grouping.py
SimonHeybrock: Hmm, good point that looks simpler, I will try.

SimonHeybrock: See what I wrote in the PR:

> Bypass the same single-threading problem encountered in SANS, I think we either have to put a helper into ESSreduce, or see if the underlying problem should be fixed in Scipp.

src/ess/dream/io/geant4.py
jl-wynen: There are no tests for the complete $I(d)$. We only test a subset of properties of the output.

src/ess/powder/grouping.py
jl-wynen: Ok. Can this be done in Scipp now instead of adding extra code here and potentially in other projects? If not, please open an issue!

SimonHeybrock: Well, the larger issue is that we have to `concat` first, which in itself is avoiding other performance issues. Making many->many mapping is a long-standing issues and has seen a couple improvements in the past, but is apparently not fully solved (See scipp/scipp#1846).

What could *probably* be done is bypass the single-threading issue, but it also required some thought (not sure there aren't any subtleties yet), but it is slightly odd to put an optimization for a solution that bypasses another problem.

jl-wynen (approved):

PR #73, opened 2024-07-08
Copier update

jl-wynen:


src/ess/powder/types.py
```diff
@@ -89,10 +90,8 @@ class DataWithScatteringCoordinates(sciline.Scope[RunType, sc.DataArray], sc.Dat
     d-spacing."""
 
 
-class NeXusDetectorDimensions(
-    sciline.Scope[NeXusDetectorName, Dict[str, int]], Dict[str, int]
-):
```
jl-wynen, line 94: This was broken because `NeXusDetectorName` is not a TypeVar. I don't see a reason why `NeXusDetectorDimensions` needs to be generic.

YooSunYoung (approved):

PR #75, opened 2024-07-10
Use ess.reduce.uncertainty

### LINKED ISSUE #37: Configurable variance handling in broadcasting
See [S.6](https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#s-6-propagation-of-uncertainties-in-broadcast-operations-should-support-drop-and-upper-bound-strategies-upper-bound-shall-be-the-default)

Implement for #25

Make use of scipp/essreduce#16.

SimonHeybrock:
Fixes #37.


src/ess/powder/types.py
```diff
@@ -56,9 +56,7 @@ class Filename(sciline.Scope[RunType, str], str):
 pixels by scattering angle into bins given by these edges.
 """
 
-UncertaintyBroadcastMode = Enum(
-    "UncertaintyBroadcastMode", ["drop", "upper_bound", "fail"]
-)
+UncertaintyBroadcastMode = _UncertaintyBroadcastMode
```
jl-wynen, line 59: How is this shown in a graph? In particular in the JSON format. Does it use `ess.powder` or `ess.reduce` as the module?

SimonHeybrock: Looks fine?
image
... but I don't know if it will still be if we include the module name?

jl-wynen: > ... but I don't know if it will still be if we include the module name?

That is my point.

But we can leave it as is for now and see how it works out.

jl-wynen (approved):

PR #78, opened 2024-08-07
Add NeXus loading helpers

### LINKED ISSUE #77: Implement NeXus loading helpers
Implement helpers that load components as well as event data individually, as in the latest versions of the ESSsans workflows.

SimonHeybrock:
This depends on a release of scipp/scippnexus#221 so CI will fail for now.

Fixes #77.

In ESSsans there is another intermediate step, adding source+sample position before adding event data, I was not sure if it makes sense the split this out here.

It is quite apparent that there is a lot of duplication of code that is in ESSsans, but currently this is required since we have per-technique `RunType` and `MonitorType`. If we could agree on a universal set it might be possible to move some of this into `essreduce`, but I am not sure for now. Or maybe someone has a better idea?


src/ess/powder/nexus.py
```diff
@@ -0,0 +1,251 @@
...
+
+Notes on the detector dimensions (2024-05-22):
+
+See https://confluence.esss.lu.se/pages/viewpage.action?pageId=462000005
+and the ICD DREAM interface specification for details.
+
+- The high-resolution and SANS detectors have a very odd numbering scheme.
+  The scheme attempts to follows some sort of physical ordering in space (x,y,z),
+  but it is not possible to reshape the data into all the logical dimensions.
+"""
```
jl-wynen, line 14: This docstring seems to be in the wrong file.

src/ess/powder/nexus.py
```diff
@@ -0,0 +1,251 @@
...
+class NXmonitor_no_events(snx.NXmonitor):
+    def __init__(
+        self, attrs: dict[str, Any], children: dict[str, snx.Field | snx.Group]
+    ):
+        children = {
+            name: child
+            for name, child in children.items()
+            if not _skip(name, child, classes=(snx.NXevent_data,))
+        }
+        super().__init__(attrs=attrs, children=children)
```
jl-wynen, line 209: To avoid duplication, you could use something like this:
```suggestion
def filtered_nx_class(base_class, exclude):
    class Filtered(base_class):
        def __init__(
                self, attrs: dict[str, Any], children: dict[str, snx.Field | snx.Group]
        ):
            children = {
                name: child
                for name, child in children.items()
                if not _skip(name, child, classes=exclude)
            }
            super().__init__(attrs=attrs, children=children)
    return Filtered
```

src/ess/powder/nexus.py
```diff
@@ -0,0 +1,251 @@
...
+    return RawSource[RunType](nexus.load_source(file_path))
+
+
+def load_nexus_detector(
+    file_path: Filename[RunType], detector_name: NeXusDetectorName
+) -> NeXusDetector[RunType]:
+    definitions = snx.base_definitions()
+    definitions["NXdetector"] = FilteredDetector
+    # Events will be loaded later. Should we set something else as data instead, or
+    # use different NeXus definitions to completely bypass the (empty) event load?
```
jl-wynen, line 68: You used this already for monitors. Why not here?

SimonHeybrock: Because ScippNexus raises when trying to use selection with a dimension or field (`event_time_zero`) which does not exist. The NXevent_data groups in the monitors of our DREAM file are completely empty. I did not see problems with the detectors, but I can double check why it works there...

jl-wynen (approved):

PR #83, opened 2024-08-26
Add sumo to endcaps in data array

celinedurniak:
This PR adds the information about the SUMOs to DREAM endcap detectors when loading csv files from GEANT4 simulations.
This additional information is required to uniquely identify their components (without relying on voxel IDs).

### DIFF
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -80,6 +80,11 @@ def _load_raw_events(file_path: str) -> sc.DataArray:
     table = sc.io.load_csv(
         file_path, sep="\t", header_parser="bracket", data_columns=[]
     )
+
+    # Add SUMO (info required for EndCaps)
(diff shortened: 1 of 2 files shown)


src/ess/dream/io/geant4.py
```diff
@@ -101,10 +106,13 @@ def group(key: str, da: sc.DataArray) -> sc.DataArray:
         if key in ["high_resolution", "sans"]:
             # Only the HR and SANS detectors have sectors.
             res = da.group("sector", *elements)
+        elif key in ["endcap_backward", "endcap_forward"]:
+            res = da.group("sumo", *elements)
```
jl-wynen, line 110: Why only for endcaps?

celinedurniak: For the other banks (mantle, HR and SANS), sumo = 1

PR #84, opened 2024-08-26
Use essreduce nexus workflow

SimonHeybrock:
The recent additions in #78 have mostly been moved into ESSreduce. We can thus remove them here, and forward to ESSreduce.

Also renamed `EmptyCanRun`->`BackgroundRun` and `EmptyInstrumentRun`->`EmptyBeamRun`. While we do not have actual scientist feedback on this I think using common names is important. I expect some future back and forth on the actual names (e.g., is `BackgroundRun` a good name?), but I hope that can be a *global* discussion across techniques, i.e., we would rename the types in ESSreduce in that case.

The Geant4 loader sees some small changes, partially relying on the providers from the NeXus workflow, but overriding some details.

Depends on an releases of scipp/essreduce#81, ~CI here will fail for now.~


src/ess/dream/io/geant4.py
```diff
@@ -167,53 +156,43 @@ def _extract_detector(
...
 
-def get_source_position(raw_source: RawSource[RunType]) -> SourcePosition[RunType]:
-    return SourcePosition[RunType](raw_source["position"])
+def geant4_load_calibration(filename: CalibrationFilename) -> CalibrationData:
+    if filename is not None:
+        # Needed to build a complete pipeline.
+        raise NotImplementedError(
+            "Calibration data loading is not implemented for DREAM GEANT4 data."
+        )
+    return CalibrationFilename(None)
```
jl-wynen, line 165:
```suggestion
    return CalibrationData(None)
```

tests/dream/geant4_reduction_test.py
```diff
@@ -47,77 +48,76 @@ def params(request):
...
+    result = workflow.compute(IofDspacingTwoTheta)
     assert result.sizes == {
         'two_theta': 16,
-        'dspacing': len(params[DspacingBins]) - 1,
+        'dspacing': len(workflow.compute(DspacingBins)) - 1,
     }
-    assert sc.identical(result.coords['dspacing'], params[DspacingBins])
-    assert sc.allclose(result.coords['two_theta'], params[TwoThetaBins])
+    assert sc.identical(result.coords['dspacing'], workflow.compute(DspacingBins))
+    assert sc.allclose(result.coords['two_theta'], workflow.compute(TwoThetaBins))
```
jl-wynen, line 112: I prefer the old implementation because if checks that the workflow respects the parameters. The new tests only check that the workflow's outputs are consistent.

SimonHeybrock: Refactored, see update.

jl-wynen (approved):

PR #90, opened 2024-09-16
Add FlatVoxelViewer for DREAM

jl-wynen:
Simple voxel diagnostics tool that can show all voxels and does not require summing or slicing.

### DIFF
--- a/docs/user-guide/dream/dream-detector-diagnostics.ipynb
+++ b/docs/user-guide/dream/dream-detector-diagnostics.ipynb
@@ -7,7 +7,7 @@
    "source": [
     "# DREAM Detector Diagnostics\n",
     "\n",
-    "This page is priomarily intended for instrument scientists and other experts.\n",
+    "This page is primarily intended for instrument scientists and other experts.\n",
     "\n",
     "ESSdiffraction has some tools for inspecting the DREAM detector based on recorded event data.\n",
     "This notebook gives an overview of the available tools.\n",
--- a/docs/user-guide/dream/dream-detector-diagnostics.ipynb
+++ b/docs/user-guide/dream/dream-detector-diagnostics.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "2dd2b714-d216-4572-96ca-f68b8935ce36",
+   "id": "0",
    "metadata": {},
    "source": [
     "# DREAM Detector Diagnostics\n",
@@ -26,7 +26,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "initial_id",
+   "id": "1",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -40,7 +40,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "05118d66-9c67-486e-9312-6518ef6308f1",
+   "id": "2",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -49,7 +49,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "10ea2f93-fd77-4b61-a585-ccdb913a0029",
+   "id": "3",
    "metadata": {},
    "source": [
     "Load simulated test data and compute histograms:"
@@ -58,7 +58,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "9f630b3a-f9f0-48ab-846b-19cdeb90f296",
+   "id": "4",
    "metadata": {},
    "outputs": [],
    "source": [
@@ -71,7 +71,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "7572f84b-e5e3-4787-9fdd-8858df0b86cd",
+   "id": "5",
    "metadata": {},
    "source": [
     "Display 2D view:"
@@ -80,7 +80,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "820d6a4e28952685",
+   "id": "6",
    "metadata": {},
    "outputs": [],
    "source": [
--- a/docs/user-guide/dream/dream-detector-diagnostics.ipynb
+++ b/docs/user-guide/dream/dream-detector-diagnostics.ipynb
@@ -1,0 +1,112 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "2dd2b714-d216-4572-96ca-f68b8935ce36",
+   "metadata": {},
+   "source": [
+    "# DREAM Detector Diagnostics\n",
+    "\n",
+    "This page is priomarily intended for instrument scientists and other experts.\n",
+    "\n",
+    "ESSdiffraction has some tools for inspecting the DREAM detector based on recorded event data.\n",
+    "This notebook gives an overview of the available tools.\n",
+    "\n",
+    "## Voxel Viewer\n",
+    "\n",
+    "The `FlatVoxelViewer` is an interactive visualization for inspecting individual voxels.\n",
+    "It shows a 2D histogram of recorded events.\n",
+    "The image axes correspond to a chosen pair of logical dimensions combined with all other dimensions.\n",
+    "Each pixel in the image corresponds to exactly one detector voxel.\n",
+    "\n",
+    "Tick marks indicate the values of the chosen dimensions.\n",
+    "E.g., all pixels between module=2 and module=3 tick marks belong to module 2."
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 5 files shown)


nvaytet:
Does the flattening scheme only apply to dream, or could we make this a generic tool for other instruments like the MaskingTool in scippneutron?

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+        if other.value == clicked.value:
+            other.value = change['old']
+        else:
+            # Only when dims are different to avoid unnecessary redraws
+            # and drawing with invalid dims.
+            for observer in self._value_observers:
+                observer(change)
+
+
+def _flat_voxel_figure(
```
nvaytet, line 144: Does the flattening scheme only apply to dream, or could we make this a generic tool for other instruments like the MaskingTool in scippneutron?

docs/user-guide/dream/dream-detector-diagnostics.ipynb
```diff
@@ -0,0 +1,112 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "2dd2b714-d216-4572-96ca-f68b8935ce36",
+   "metadata": {},
+   "source": [
+    "# DREAM Detector Diagnostics\n",
+    "\n",
+    "This page is priomarily intended for instrument scientists and other experts.\n",
```
nvaytet, line 10:
```suggestion
    "This page is primarily intended for instrument scientists and other experts.\n",
```

docs/user-guide/dream/index.md
```diff
@@ -7,4 +7,5 @@ maxdepth: 1
 
 dream-data-reduction
 dream-instrument-view
+cream-detector-diagnostics
```
nvaytet, line 10:
```suggestion
dream-detector-diagnostics
```

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
```
nvaytet, line 2:
```suggestion
# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
```

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+import numpy as np
+import scipp as sc
+from plopp.core.typing import FigureLike
+
+# This leads to y-axis: segment, x-axis: module
+_STARTING_DIMS = ('segment', 'wire', 'strip', 'module', 'counter', 'sector')
+
+
+class FlatVoxelViewer(ipw.VBox):
+    def __init__(self, data: sc.DataGroup) -> None:
```
nvaytet, line 19: Add docstring?

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+        return self._horizontal_selector_box.children[0]
+
+    @property
+    def _vertical_buttons(self) -> ipw.ToggleButtons:
+        return self._vertical_selector_box.children[0]
+
+    def set_dims(self, new_dims: tuple[str, ...]) -> None:
+        self._horizontal_buttons.unobserve_all(name='value')
+        self._vertical_buttons.unobserve_all(name='value')
+        # self._forward_messages = False
```
nvaytet, line 99: remove commented code

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+        return {
+            'horizontal': self._horizontal_buttons.value,
+            'vertical': self._vertical_buttons.value,
+        }
+
+    def observe(self, handler, names: str | list[str], *args, **kwargs):
+        if names != 'value':
+            super().observe(handler, names, *args, **kwargs)
+        else:
+            self._value_observers.append(handler)
```
nvaytet, line 126: Not sure I understand the mechanism with the `value_observers`. Maybe you can explain to me?

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+            # Only when dims are different to avoid unnecessary redraws
+            # and drawing with invalid dims.
+            for observer in self._value_observers:
+                observer(change)
+
+
+def _flat_voxel_figure(
+    data: sc.DataArray, horizontal_dim: str, vertical_dim: str
+) -> FigureLike:
+    # flat, h_coord, v_coord = _flatten_voxel_dims(data, horizontal_dim, vertical_dim)
```
nvaytet, line 147: Commented code

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+    # This relies on the order of flatten_to_h/v
+    inner_volume_h = _product(data.sizes[d] for d in flatten_to_h[1:])
+    inner_volume_v = _product(data.sizes[d] for d in flatten_to_v[1:])
+    h_ticks = np.arange(0, flat.sizes['horizontal'], inner_volume_h)
+    v_ticks = np.arange(0, flat.sizes['vertical'], inner_volume_v)
+
+    h_labels = [str(value) for value in h_coord.values]
+    v_labels = [str(value) for value in v_coord.values]
+
+    fig = flat.plot(rasterized=True)
```
nvaytet, line 178: I would probably make `rasterized` a kwarg for creating the viewer? (with `True` as default).

src/ess/dream/diagnostics.py
```diff
@@ -0,0 +1,228 @@
...
+
+    fig.ax.xaxis.set_ticks(ticks=h_ticks, labels=h_labels)
+    fig.ax.yaxis.set_ticks(ticks=v_ticks, labels=v_labels)
+    fig.ax.set_xlabel(horizontal_dim.capitalize())
+    fig.ax.set_ylabel(vertical_dim.capitalize())
```
nvaytet, line 183:
```suggestion
    fig.canvas.xlabel(horizontal_dim.capitalize())
    fig.canvas.ylabel(vertical_dim.capitalize())
```

src/ess/dream/diagnostics.py
jl-wynen: It can be generalised easily. Which other instruments have multi-logical-dimensional detectors?

docs/user-guide/dream/index.md
jl-wynen: I might have been hungry :)

src/ess/dream/diagnostics.py
jl-wynen: This is about updating only once when dimensions are swapped between x and y. I tried to find a way to stop widgets from emitting messages. But I couldn't get this to work. So instead, I put a custom layer in between so that I can control what messages make it to the user.

All observers with `names='value` are registered in `self._value_observers`. And the actual handler for the toggle buttons is `self._observe_values`. This function calls the user handlers when appropriate.

src/ess/dream/diagnostics.py
nvaytet: Most of them? I remember when Morten was talking about the EFU, all instruments seems to have some logical ordering. Loki for sure has banks, tubes, straws, pixels.

src/ess/dream/diagnostics.py
nvaytet: How about having a lock?
Add a `self._lock = False`.

Then in the function that gets called on update:
```Py
def update(change):
    if self._lock:
        return
    if other.value == clicked.value:
        self._lock = True
        other.value = change["old"]
        self._lock = False
    ...
```
Would that work? It would simplify a lot of the logic I think?

src/ess/dream/diagnostics.py
jl-wynen: I would suggest, we merge this here and ask for feedback. If that is positive, we can generalise the viewer and move it into ESSreduce. What do you think?

jl-wynen:
The code may need to be updated when #92 is merged. So I propose to wait for that PR.

src/ess/dream/diagnostics.py
jl-wynen: I don't think this would work. This would prevent updating within `_DimensionSelector` while the lock is held. But the observers of the button that control the actual figure would still get called on `other.value = change["old"]`.

jl-wynen: Seems to work. Thanks!

I don't think this solution is valid for multi-threaded applications. But I am not 100% confident about my old implementation either. And it should be fine because ipywidgets does not spawn threads, right?

nvaytet (approved):

PR #92, opened 2024-09-18
Load SUMO coordinate from GEANT4 CSV files

jl-wynen:
Based on an implementation by @celinedurniak, see also #83.

### DIFF
--- a/src/ess/dream/io/geant4.py
+++ b/src/ess/dream/io/geant4.py
@@ -69,6 +69,7 @@ def _load_raw_events(file_path: str) -> sc.DataArray:
     table = sc.io.load_csv(
         file_path, sep="\t", header_parser="bracket", data_columns=[]
     )
+    table.coords['sumo'] = table.coords['det ID']
     table.coords.pop("lambda", None)
     table = table.rename_dims(row="event")
     return sc.DataArray(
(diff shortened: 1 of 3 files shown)


src/ess/dream/io/geant4.py
```diff
@@ -90,10 +91,14 @@ def group(key: str, da: sc.DataArray) -> sc.DataArray:
         if key in ["high_resolution", "sans"]:
             # Only the HR and SANS detectors have sectors.
             res = da.group("sector", *elements)
+        elif key in ["endcap_backward", "endcap_forward"]:
+            # Other banks only have a single SUMO.
+            res = da.group("sumo", *elements)
```
YooSunYoung, line 96: Even if other banks only have a single SUMO, we can still group them by `SUMO`, no...?

YooSunYoung (approved):

src/ess/dream/io/geant4.py
jl-wynen: We can. But that would be the use? This extra dimension would get in the way of plotting and would always have to be sliced out.

YooSunYoung: Yeah, I wasn't sure how many times we have to filter the endcap for this reason.
If it's not in many places, I think not-grouping is fine.

PR #93, opened 2024-09-18
DREAM monitor normalization

### LINKED ISSUE #76: [Requirement]  Normalise by monitor
### Executive summary

Normalise data by monitor counts

This issue is an alternative to #30. In some cases users want to normalise detector data by a monitor. This can be either
- normalization by a wavelength spectrum of the monitor,
- normalization by an integrated monitor (single number).

The user needs to be able to pick one or the other.

### Inputs

- Data array for detector counts in wavelength
- Data array for monitor counts
  - In wavelength if normalising by wavelength spectrum.
  - Arbitrary coordinate when normalising by integrated monitor.

### Methodology

*Normalisation by wavelength spectrum*
Either histogram the monitor into the same bins as the detector data and then divide. Or use `sc.lookup` to account for different binning. (See old WISH workflow.)

*Normalisation by integrated monitor*
Divide detector data by `sc.sum(monitor)`.

### Outputs

Normalised data array.

We should probably ensure that the data array has the same units regardless of which monitor normalisation was used.

Integrated into reduction workflow, Python module / function

### Test cases

Dummy test data.

### Comments

_No response_

jl-wynen:
Fixes #76

### DIFF
--- a/src/ess/powder/correction.py
+++ b/src/ess/powder/correction.py
@@ -107,7 +107,7 @@ def normalize_by_monitor_integrated(
     lo = det_coord.min()
     hi = det_coord.max()
     monitor = monitor[dim, lo:hi]
-    # Strictly limit `monitor` t the range of `detector`.
+    # Strictly limit `monitor` to the range of `detector`.
     edges = sc.concat([lo, monitor.coords[dim][1:-1], hi], dim=dim)
     monitor = sc.rebin(monitor, {dim: edges})
 


src/ess/dream/workflow.py
```diff
@@ -37,13 +40,14 @@ def default_parameters() -> dict:
...
 
-def DreamGeant4Workflow() -> sciline.Pipeline:
+def DreamGeant4Workflow(*, run_norm: RunNormalization) -> sciline.Pipeline:
     """
     Workflow with default parameters for the Dream Geant4 simulation.
     """
     wf = LoadGeant4Workflow()
     for provider in itertools.chain(powder_providers, _dream_providers):
         wf.insert(provider)
+    run_norm.insert(wf)
```
nvaytet, line 50: I was quite confused by this interface. I actually thought you got the order wrong, and that it should have been `wf.insert(run_norm)`.
Then I saw that you added a `insert` method to the `Enum`.

I was expecting this to work more like e.g. the `UncertaintyBroadcastMode`, which is just a parameter we set on the workflow (e.g. https://github.com/scipp/esssans/blob/main/docs/user-guide/isis/sans2d.ipynb?short_path=e98bbec#L141)?

Was there a reason I'm missing that meant you chose this interface?

src/ess/powder/correction.py
```diff
@@ -2,81 +2,113 @@
...
     ----------
-    data:
-        Input event data.
+    detector:
+        Input event data in wavelength.
     monitor:
-        A histogrammed monitor.
-    wavelength_edges:
-        If given, rebin the monitor with these edges.
-    smooth_args:
```
nvaytet, line 50: So you removed the smoothing of the monitor. Was that intentional?

src/ess/powder/correction.py
```diff
@@ -2,81 +2,113 @@
...
-        monitor = lowpass(monitor, dim="wavelength", **smooth_args)
-    return data.bins / sc.lookup(func=monitor, dim="wavelength")
+    # Clip `monitor` to the range of `detector`.
+    det_coord = (
+        detector.coords[dim] if dim in detector.coords else detector.bins.coords[dim]
+    )
+    lo = det_coord.min()
+    hi = det_coord.max()
+    hi.value = np.nextafter(hi.value, np.inf)
+    monitor = monitor[dim, lo:hi]
```
nvaytet, line 104: What happens if the monitor spans a narrower range than the detector?

src/ess/powder/correction.py
```diff
@@ -2,81 +2,113 @@
...
+
+    coord = monitor.coords[dim]
+    norm = sc.sum(monitor.data * (coord[1:] - coord[:-1]))
+    norm = broadcast_uncertainties(
+        norm, prototype=detector, mode=uncertainty_broadcast_mode
+    )
+    return NormalizedRunData[RunType](detector / norm)
```
nvaytet, line 111: So you don't need a `sc.lookup` here because you are broadcasting to all events, is that correct?

src/ess/powder/conversion.py
```diff
@@ -175,7 +178,7 @@ def _restore_tof_if_in_wavelength(data: sc.DataArray) -> sc.DataArray:
 
 
 def add_scattering_coordinates_from_positions(
-    data: NormalizedByProtonCharge[RunType], graph: ElasticCoordTransformGraph
```
nvaytet, line 178: I'm not sure I managed to follow why this is now `FilteredData`... I would have expected `NormalizedRunData`?

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -285,6 +285,124 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "26",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "workflow[MonitorFilename[SampleRun]] = dream.data.simulated_monitor_diamond_sample()\n",
+    "workflow[MonitorFilename[VanadiumRun]] = dream.data.simulated_monitor_vanadium_sample()\n",
+    "workflow[MonitorFilename[BackgroundRun]] = dream.data.simulated_monitor_empty_can()\n",
+    "workflow[CaveMonitorPosition] = sc.vector([0.0, 0.0, -4220.0], unit='mm')\n",
```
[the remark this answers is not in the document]
jl-wynen: Correct.

src/ess/dream/workflow.py
jl-wynen: 1. I used `run_norm.insert(wf)` instead of `wf.insert(run_norm)` because I thought it would be good to keep the selection logic within `RunNormalization`. I see how the call can be confusing. I can change it to a free function. But I don't want the `match` statement inlined into `DreamGeant4Workflow`.
2. We cannot use a parameter like `UncertaintyBroadcastMode` because the different normalizations need to request different data from the pipeline. Using a combined providers that picks a norm would require requesting the union of all dependencies, that is monitors *and* proton charge.

src/ess/powder/correction.py
jl-wynen: Yes. This should be done as a separate pipeline step if desired.

src/ess/powder/correction.py
jl-wynen: Then the normalization is off, I think.

What do you think should happen?
1. Keep it as is.
2. Raise an error.
3. Crop the data to the monitor range.

This also affects `normalize_by_monitor_histogram` which currently just uses whatever values lookup produces.

src/ess/powder/correction.py
jl-wynen: Yes. And because `norm` is a scalar.

src/ess/powder/conversion.py
jl-wynen: Monitor normalization needs a wavelength coord. So it has to be done after `add_scattering_coordinates_from_positions`. For proton charge normalization, the order does not matter. I moved it to the same place as monitor normalization to make it easier to build the graph.

src/ess/dream/workflow.py
nvaytet: I think it would be better to have a free function and remove it from the Enum?

An alternative could be to make classmethods that would generate the workflow with the correct providers?
e.g.
```Py
LoadGeant4Workflow.normalized_by_monitor_histogram()
LoadGeant4Workflow.normalized_by_monitor_integrated()
LoadGeant4Workflow.normalized_by_proton_charge()
```
and one of them could be the default?

nvaytet: Ah yes, I forgot that it wasn't a class :roll_eyes:

src/ess/powder/correction.py
jl-wynen: Done

src/ess/dream/workflow.py
jl-wynen: Done

src/ess/powder/correction.py
```diff
@@ -2,81 +2,135 @@
...
+    # Clip `monitor` to the range of `detector`.
+    det_coord = (
+        detector.coords[dim] if dim in detector.coords else detector.bins.coords[dim]
+    )
+    lo = det_coord.min()
+    hi = det_coord.max()
+    # hi is shifted towards MINUS infinity because label-based indexing with bin-edges
+    # in inclusive for the upper edge. But we don't want the literal upper edge
+    # of the detector to be included.
+    hi.value = np.nextafter(hi.value, -np.inf)
```
jl-wynen, line 112: This was broken in my first implementation. I used `np.nextafter(hi.value, +np.inf)`. Please check that the logic is sound.

src/ess/powder/types.py
```diff
@@ -28,6 +29,7 @@
 EmptyBeamRun = reduce_gt.EmptyBeamRun
 Filename = reduce_gt.Filename
 Incident = reduce_gt.Incident
+CaveMonitor = reduce_gt.Monitor1
```
SimonHeybrock, line 32: Have you considering naming this by what is does instead of where it is?

src/ess/powder/correction.py
```diff
@@ -260,6 +318,27 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
...
+    workflow: sciline.Pipeline, run_norm: RunNormalization
+) -> None:
+    """Insert providers for a specific normalization into a workflow."""
+    match run_norm:
+        case RunNormalization.monitor_histogram:
+            workflow.insert(normalize_by_monitor_histogram)
+        case RunNormalization.monitor_integrated:
+            workflow.insert(normalize_by_monitor_integrated)
+        case RunNormalization.proton_charge:
+            workflow.insert(normalize_by_proton_charge)
```
SimonHeybrock, line 339: Maybe I lack some context, but I am not convinced by the design choice here. Why do we need an enum which then decides which providers to insert? Wouldn't it be more natural and a more modular design if we had a small normalization workflow (or 3 actually) which can be passed to (dependency injected into) the main workflow creation?

jl-wynen: Yes, that would be great. But it requires https://github.com/scipp/sciline/issues/180

jl-wynen: If we could do this, would you require the users to know where to insert the workflow? I.e.
```python
wf = DreamWorkflow()
norm = MonitorHistogramNormalization()
wf[NormalizedRunData[RunType]] = norm
```
Or do we still need a helper function?
```python
wf = DreamWorkflow()
norm = MonitorHistogramNormalization()
insert_run_normalization(wf, norm)
```

SimonHeybrock: Neither:
```python
wf = DreamWorkflow(normalization_workflow=MonitorHistogramNormalization())
```
or something like that.

We should look into scipp/sciline#180 then, before adding workarounds.

SimonHeybrock: As upstream changes ran into trouble, I suggest we keep this as it is.

src/ess/powder/correction.py
```diff
@@ -2,81 +2,137 @@
...
-            smooth_args,
+    # Clip `monitor` to the range of `detector`, where the bins at the boundary
+    # may extend past the detector range (how label-based indexing works).
+    det_coord = (
+        detector.coords[dim] if dim in detector.coords else detector.bins.coords[dim]
+    )
+    lo = det_coord.min()
+    hi = det_coord.max()
+    monitor = monitor[dim, lo:hi]
+    # Strictly limit `monitor` t the range of `detector`.
```
nvaytet, line 110:
```suggestion
    # Strictly limit `monitor` to the range of `detector`.
```

nvaytet (approved):

src/ess/powder/correction.py
nvaytet: Conclusion from offline discussion was: `rebin` actually works.

PR #94, opened 2024-09-25
Remove unused domain types

SimonHeybrock:
@celinedurniak were there others that are not needed that you had in mind?


celinedurniak:
What are `RunType` and `ScatteringRunType`?

SimonHeybrock:
> What are `RunType` and `ScatteringRunType`?

These a type-variables, which will (in the case of powder) take values such as `SampleRun` or `VanadiumRun`. Any data-reduction step that performs identical computation for several run types uses these. Here is an example:

https://github.com/scipp/essdiffraction/blob/7772e6962b7739da6c74e2b809d90d278b61ee58/src/ess/powder/conversion.py#L199-L203

This function will/can be used multiple times in the reduction workflow graph, typically for sample and vanadium runs.

celinedurniak:
And 'NexusSource' is the input raw file?

SimonHeybrock:
> And 'NexusSource' is the input raw file?

It is the NXsource group loaded from a Nexus file. See https://scipp.github.io/essreduce/generated/modules/ess.reduce.nexus.generic_types.html for full list of types and explanations.

celinedurniak:
Thanks for the explanations. 
So, to answer the first question, at the moment, I do not see any more terms to remove.

PR #96, opened 2024-10-04
Workflow widget interface

jl-wynen:
An adapted version of https://github.com/scipp/esssans/pull/171. That PR should be finalised first and this PR should then be updated to match https://github.com/scipp/esssans/pull/171.


jl-wynen:
This currently fails because something is wrong with the pixel mask filenames. I don't understand why it doesn't work. It seems to do the same as in the loki workflow. Except that there are no massks here.

nvaytet:
@jl-wynen should we try to fix this before the STAPs?

jl-wynen:
I'm not allowed to approve. But it looks good to me.

nvaytet (approved):

PR #99, opened 2024-10-10
Update for restructured NeXus base workflow from `ess.reduce.nexus`

SimonHeybrock:
Required when scipp/essreduce#114 is released.


nvaytet (approved):
Update deps?

SimonHeybrock:
> Update deps?

Not merged and released yet, i.e., not possible.

PR #104, opened 2024-11-08
feat: absorption correction

### LINKED ISSUE #22: Absorption correction for cylindrical samples
### Executive summary

Calculate and apply absorption correction for cylindrical samples in Powder Diffraction

For example, see algorithms in Mantid

[AbsorptionCorrection](https://docs.mantidproject.org/nightly/algorithms/AbsorptionCorrection-v1.html)

[CylinderAbsorption](https://docs.mantidproject.org/nightly/algorithms/CylinderAbsorption-v1.html)

[SNSPowderReduction](https://docs.mantidproject.org/nightly/algorithms/SNSPowderReduction-v1.html?highlight=snspowderreduction) in Mantid to check when the absorption correction is applied within the whole workflow

### Inputs

Scipp data structure from loaded NeXus or csv file(s)

**Assumption:** the input should be expressed in wavelength

Note that information about the sample (chemical formula, geometry, mass density, number density...) might be stored in the sample log of the file or must have to be provided by the user when using the algorithm.

### Methodology

See source code from (Mantid) references above

**Assumption:**  we consider only cylindrical samples for this first implementation

**Acceptance criterion** 
Absorption correction step is implemented in ESS Diffraction worflow (https://scipp.github.io/ess/instruments/external/powgen/powgen.html)

### Outputs

This requirement should provide the absorption correction term and the corrected sample data

Integrated into reduction workflow, Python module / function

### Test cases

- As input file, use POWGEN event file provided in Mantid (PG3_4866_event.nxs or PG3_4844_event.nxs) and already used in ESS Diffraction workflow.
In this case, the output from the implementation in Scipp could be compared with what we get from Mantid.
- CSV files from DREAM (also available in Scipp documentation)

### Comments

**Followup Work:** Extend this implementation to:

 -  other sample shapes
 -  several components (sample, container)
 -  using other assumptions (like ‘FullPaalmanPings’)

jokasimr:
Fixes #22 

This is a draft, but I need some help here so please take a look anyway and tell me if there's a better way.

The issue I'm not happy with here is the re-binning of the events.
Essentially what we want to do is to `lookup` the absorption correction from `position` and `wavelength` for each event.

### DIFF
--- a/src/ess/powder/conversion.py
+++ b/src/ess/powder/conversion.py
@@ -15,7 +15,7 @@ from .types import (
     DspacingData,
     ElasticCoordTransformGraph,
     MaskedData,
-    NormalizedByProtonCharge,
+    NormalizedByAbsorption,
     RunType,
 )
 
@@ -175,7 +175,7 @@ def _restore_tof_if_in_wavelength(data: sc.DataArray) -> sc.DataArray:
 
 
 def add_scattering_coordinates_from_positions(
-    data: NormalizedByProtonCharge[RunType], graph: ElasticCoordTransformGraph
+    data: NormalizedByAbsorption[RunType], graph: ElasticCoordTransformGraph
 ) -> DataWithScatteringCoordinates[RunType]:
     """
     Add ``wavelength`` and ``two_theta`` coordinates to the data.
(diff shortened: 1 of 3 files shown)


jokasimr:
Celine has verified that it's fine to compute the correction not for every pixel but for a coarser grid containing the detector pixels.

PR #106, opened 2024-11-14
Convert reduced data back to tof

### LINKED ISSUE #97: Save data in TOF to CIF
We need to write the final results in TOF so that analysis software can read them. In the CIF writer, this means changing from saving `pd_proc` to `pr_meas` with the appropriate fields.

jl-wynen:
Fixes #97

### DIFF
--- a/docs/user-guide/dream/dream-data-reduction.ipynb
+++ b/docs/user-guide/dream/dream-data-reduction.ipynb
@@ -20,8 +20,6 @@
    "outputs": [],
    "source": [
     "import scipp as sc\n",
-    "import scippneutron as scn\n",
-    "import scippneutron.io\n",
     "\n",
     "from ess import dream, powder\n",
     "import ess.dream.data  # noqa: F401\n",
@@ -90,7 +88,7 @@
    "source": [
     "## Use the workflow\n",
     "\n",
-    "We can visualize the graph for computing the final normalized result for intensity as a function of d-spacing:"
+    "We can visualize the graph for computing the final normalized result for intensity as a function of time-of-flight:"
    ]
   },
   {
@@ -100,7 +98,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "workflow.visualize(IofDspacing, graph_attr={\"rankdir\": \"LR\"})"
+    "workflow.visualize([IofTof, ReducedTofCIF], graph_attr={\"rankdir\": \"LR\"})"
    ]
   },
   {
@@ -108,7 +106,8 @@
    "id": "8",
    "metadata": {},
    "source": [
-    "We then call `compute()` to compute the result:"
+    "We then call `compute()` to compute the result:\n",
+    "(The `cif` object will later be used to write the result to disk.)"
    ]
   },
   {
@@ -118,8 +117,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "result = workflow.compute(IofDspacing)\n",
-    "result"
+    "results = workflow.compute([IofTof, ReducedTofCIF])\n",
+    "result = results[IofTof]\n",
+    "cif_data = results[ReducedTofCIF]"
    ]
   },
   {
@@ -129,8 +129,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dspacing_histogram = result.hist()\n",
-    "dspacing_histogram.plot()"
+    "histogram = result.hist()\n",
+    "histogram.plot()"
    ]
   },
   {
@@ -148,11 +148,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "to_save = dspacing_histogram.assign_coords(\n",
-    "    dspacing=sc.midpoints(\n",
-    "        dspacing_histogram.coords[\"dspacing\"]\n",
-    "    ))\n",
-    "scn.io.save_xye(\"dspacing.xye\", to_save)"
+    "cif_data.comment = \"\"\"This file was generated with the DREAM data reduction user guide\n",
+    "in the documentation of ESSdiffraction.\n",
+    "See https://scipp.github.io/essdiffraction/\n",
+    "\"\"\"\n",
+    "cif_data.save('reduced.cif')"
    ]
   },
   {
@@ -333,7 +333,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "workflow.visualize(IofDspacing, graph_attr={\"rankdir\": \"LR\"})"
+    "workflow.visualize(IofTof, graph_attr={\"rankdir\": \"LR\"})"
    ]
   },
   {
@@ -343,8 +343,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "results = workflow.compute((IofDspacing, WavelengthMonitor[SampleRun, CaveMonitor]))\n",
-    "normalized_by_monitor = results[IofDspacing]\n",
+    "results = workflow.compute((IofTof, WavelengthMonitor[SampleRun, CaveMonitor]))\n",
+    "normalized_by_monitor = results[IofTof]\n",
     "monitor = results[WavelengthMonitor[SampleRun, CaveMonitor]]\n",
     "monitor"
    ]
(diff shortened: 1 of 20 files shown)


src/ess/powder/conversion.py
```diff
@@ -229,8 +269,10 @@ def convert_monitor_do_wavelength(
 
 
 providers = (
+    assemble_output_calibration,
     powder_coordinate_transformation_graph,
     add_scattering_coordinates_from_positions,
     convert_to_dspacing,
+    convert_to_calibrated_tof,
```
celinedurniak, line 276: 'calibrated' might be misleading. It could be replaced by 'reduced', 'derived', 'final'

docs/user-guide/dream/dream-data-reduction.ipynb
```diff
@@ -148,11 +148,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "to_save = dspacing_histogram.assign_coords(\n",
-    "    dspacing=sc.midpoints(\n",
-    "        dspacing_histogram.coords[\"dspacing\"]\n",
-    "    ))\n",
-    "scn.io.save_xye(\"dspacing.xye\", to_save)"
+    "cif_data.comment = \"\"\"This file was generated with the DREAM data reduction user guide\n",
```
celinedurniak, line 151: Additional metadata should be added (if not stored somewhere else in the CIF file):
- version of essdiffraction used
- date of creation
- instrument name and detector bank
- sample name
- raw files: sample and vanadium (if used)
- values and units of average 2theta, average Ltotal, dtt1 or difc

src/ess/powder/conversion.py
```diff
@@ -211,11 +218,44 @@ def convert_to_dspacing(
         for key in ('wavelength', 'two_theta'):
             if key in out.coords.keys():
                 out.coords.set_aligned(key, False)
-    out.bins.coords.pop('tof', None)
+    # out.bins.coords.pop('tof', None)
```
SimonHeybrock, line 221: Commented code

src/ess/powder/conversion.py
```diff
@@ -211,11 +218,44 @@ def convert_to_dspacing(
...
+        2
+        * sc.constants.m_n
+        / sc.constants.h
+        * average_l
+        * sc.sin(0.5 * average_two_theta),
+        unit='us / angstrom',
+    ).value
+    return OutputCalibrationData(
+        sc.DataArray(
+            sc.array(dims=['calibration'], values=[difc]),
```
SimonHeybrock, line 242: Why are why taking the value (stripping units?) and put it back into a variable?

src/ess/powder/conversion.py
```diff
@@ -211,11 +218,44 @@ def convert_to_dspacing(
...
+            coords={'power': sc.array(dims=['calibration'], values=[1])},
+        )
+    )
+
+
+def convert_to_calibrated_tof(
+    data: IofDspacing, calibration: OutputCalibrationData
+) -> IofTof:
+    difc = calibration['power', sc.scalar(1)].data
```
SimonHeybrock, line 251: Is the plan to have other powers as well? Is a `DataArray` the right data structure here? Maybe a `DataGroup` or simple `dataclass` would be more intuitive?

docs/user-guide/dream/dream-data-reduction.ipynb
jl-wynen: I agree. But that should be a separate PR. In particular because we don't even know what some metadata (esp. sample) looks like in the input.

But note that this is filled in automatically:
```
_audit.creation_date 2024-11-14T16:20:10+00:00
_audit.creation_method 'Written by scippneutron v24.11.0'
_computing.diffrn_reduction 'ess.dream v24.10.1.dev32+gacad30d.d20241114'

_diffrn_radiation.probe neutron
_diffrn_source.beamline DREAM
_diffrn_source.facility ESS
_diffrn_source.device spallation
```

src/ess/powder/conversion.py
jl-wynen: Because the underlying CIF writer wants the coefficients as a data array (https://scipp.github.io/scippneutron/generated/modules/scippneutron.io.cif.CIF.html#scippneutron.io.cif.CIF.with_powder_calibration):
```python
cal = sc.DataArray(
    sc.array(dims=['cal'], values=[3.4, 0.2]),
    coords={'power': sc.array(dims=['cal'], values=[0, 1])},
)
```
It is unfortunate that we have to remove the unit but that is how CIF wants the data.

We could instead keep a dict until the last moment where we need to convert to the above data array:
```python
{
   0: sc.scalar(3.4, unit='us'),
   1: sc.scalar(0.2, unit='us/angstrom')
}
```
Would you prefer this?

src/ess/powder/conversion.py
jl-wynen: > Is the plan to have other powers as well?

Ultimately, yes, as far as I know. @celinedurniak ?

See https://github.com/scipp/essdiffraction/pull/106#discussion_r1843367827 why it is a data array. A data group would not work because the keys should be powers (`int`) and not names (`str`). A dataclass would have a fixed set of fields. Plus it would also need names instead of powers.

src/ess/powder/conversion.py
```diff
@@ -211,11 +218,44 @@ def convert_to_dspacing(
...
+    difc.unit = 'us / angstrom'
+    res = data.drop_coords('dspacing').bins.drop_coords('dspacing')
+    res.coords['tof'] = sc.to_unit(difc * data.coords['dspacing'], unit='us')
+    res.bins.coords['tof'] = sc.to_unit(difc * data.bins.coords['dspacing'], unit='us')
```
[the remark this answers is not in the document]
jl-wynen: I hadn't. Becuase the callibration coefficients are not coordinates. I could use a closure to capture them as a transformation function?

src/ess/powder/conversion.py
SimonHeybrock: I don't mind much, but can you explain what is going on in the comments?

src/ess/powder/conversion.py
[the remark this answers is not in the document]
SimonHeybrock: It would avoid having to manually handle dense and binned coords. Give it a try, but this is also fine?

src/ess/powder/conversion.py
jl-wynen: Better now?

src/ess/powder/conversion.py
jl-wynen: Have a look at the new implementation.

src/ess/powder/conversion.py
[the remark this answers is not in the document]
jl-wynen: See update

src/ess/powder/calibration.py
```diff
@@ -0,0 +1,118 @@
...
+import scipp as sc
+import scipp.constants
+
+from .types import DspacingData, SampleRun
+
+
+class ScalarCalibrationData(Mapping[int, sc.Variable]):
+    r"""Calibration data for output ToF data.
+
+    Only of value is stored per coefficient.
```
SimonHeybrock, line 19: grammar

src/ess/powder/calibration.py
```diff
@@ -0,0 +1,118 @@
...
+        cal = self.to_cif_units()
+        return sc.DataArray(
+            sc.array(
+                dims=['calibration'], values=[c.value for c in cal.values()], unit=None
+            ),
+            coords={'power': sc.array(dims=['calibration'], values=list(cal.keys()))},
+        )
+
+
+OutputCalibrationData = NewType('OutputCalibrationData', ScalarCalibrationData)
```
SimonHeybrock, line 97: Can there be different ones? Why do we need the alias?

jl-wynen: I guess not

src/ess/powder/conversion.py
celinedurniak: @jl-wynen your assumption is correct. There could be, for example, these terms to consider
difc * dspacing + difa * dispacing **2 + difb/dspacing +Zero

jl-wynen: Ok. Do we need to support them now or can this wait until we actually need them?

SimonHeybrock (approved):

PR #113, opened 2024-12-10
Add illustration of inhomogeneity from absorption

jokasimr:
@celinedurniak mentioned that for #22 it would be good to be able to visualize the effect of absorption on the intensity in the detector.


jokasimr:
Screenshot from 2024-12-10 13-18-01

docs/user-guide/dream/dream-instrument-view.ipynb
```diff
@@ -229,6 +229,65 @@
    "source": [
     "dream.instrument_view(dg[\"endcap_backward\"][\"module\", 0].hist(tof=1))"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "15",
+   "metadata": {},
+   "source": [
+    "## Display inhomogeneity from absorption\n",
```
nvaytet, line 238: Nice way of visualizing, but I am not sure this is the right notebook to put it in?
It feels sort of unrelated to the rest?

Would it be better as part of something like a gallery notebook (even though we don't have a gallery yet?).
Or even keep a list of 'recipes' in e.g. our github discussions (and refer to them in the docs)?
This visualization sounds like it is something that the instrument scientist would find useful just for themselves, but isn't a necessary part of a reduction workflow?

jokasimr: Sure, we can put it in a separate notebook.

It's mainly useful for IS to estimate what kind of impact absorption has in some experiment setup.
Celine mentioned that this would be good step towards having an automatic absorption correction.
Making absorption correction part of the workflow is started in https://github.com/scipp/essdiffraction/pull/104/

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,140 @@
...
+    "# DREAM instrument view\n",
+    "\n",
+    "This notebook is a simple example of how to use the instrument view for the DREAM instrument.\n",
+    "\n",
+    "- The DREAM-specific instrument view is capable of slicing the data with a slider widget along a dimension (e.g. `tof`) by using the `dim` argument.\n",
+    "- There are also checkboxes to hide/show the different elements that make up the DREAM detectors.\n",
+    "\n",
+    "We begin with relevant imports.\n",
+    "We will be using tutorial data downloaded with `pooch`.\n",
+    "If you get an error about a missing module `pooch`, you can install it with `!pip install pooch`:"
```
nvaytet, line 17: I think the text in here needs updating? (seems it was copied from the instrument view notebook)

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,171 @@
+{
```
nvaytet, line 1: If you want this notebook to appear in the docs, you need to add it to the `dream/index.md`.

jokasimr: Forgot that, thanks.

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -140,10 +196,10 @@
   },
   {
    "cell_type": "markdown",
-   "id": "9",
+   "id": "13",
    "metadata": {},
    "source": [
-    "As expected the impact of absorption on the inhomgenity of the intensity is not nearly as large for the endcap detector as it is for the mantle detector."
+    "As expected the impact of absorption on on the intensity in the endcap detector is close to uniform."
```
nvaytet, line 202: It doesn't look so uniform when you look at the figure... until you look at the colorbar.
Screenshot_20250123_142736

Should we use some common colorbar limits for all 3 views?

Also, do we really have to use the heavy instrument view, or could we use 2d scatter plots instead?
You could make z-y plots for the mantle, and x-y plot for the end cap?

The 3d views dramatically increase the size of the docs, and thus the repository itself.
It would also mean that you don't need to mess around with Camera settings...

jokasimr: > Should we use some common colorbar limits for all 3 views?

No I don't think that's important. It's better to keep it simple and to make it look like something they might encounter in their own notebook.

> The 3d views dramatically increase the size of the docs, and thus the repository itself.

I didn't know that, why does it increase the size of the repo.
We could make it 2d, but the 3d view is nicer to play around with if you actually run the notebook.

jokasimr: Took a look at the sizes now, after building the docs locally I have the following files:
```
-rw-rw-r-- 1 johannes johannes  17M jan 23 09:26 dream-instrument-view.html
-rw-rw-r-- 1 johannes johannes  18M jan 23 09:25 dream-instrument-view.ipynb
-rw-rw-r-- 1 johannes johannes 4,5M jan 23 09:26 dream-visualize-absorption.html
-rw-rw-r-- 1 johannes johannes 4,8M jan 23 09:25 dream-visualize-absorption.ipynb
```
So it seems the absorption visualization notebook is significantly smaller than the instrument view notebook.

I also tried cloning `essdiffraction` to check how fast it was to download:
```
~/crap $ git clone https://github.com/scipp/essdiffraction.git
Cloning into 'essdiffraction'...
Receiving objects: 100% (3761/3761), 9.63 MiB | 10.66 MiB/s, done.
```
Compressed it seems the size is closer to 10Mb.

Overall I get the impression that it doesn't really matter if we put this file here, it's unlikely to make the repo hard to download for anyone.

nvaytet: 10Mb is a lot. Maybe we should really reduce the size of the instrument view notebook?

> the 3d view is nicer to play around with.

It might be nice to play with, but I don't think it adds any information to what we are viewing? I think the views might be clearer in 2D. I will have a go at making them and we can compare.

jokasimr: 1. 10Mb is not a lot, it takes 2s to download.
2. This change is easily revertible. If we notice slowness we can update the docs and redeploy.

> It might be nice to play with, but I don't think it adds any information to what we are viewing?

Let's just ask Celine what she prefers? Then we don't have to guess.

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,227 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "3",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dg = dream.io.load_geant4_csv(dream.data.get_path(\"data_dream0_new_hkl_Si_pwd.csv.zip\"))[\"instrument\"]\n",
+    "dg = sc.DataGroup({key: detector[\"events\"] for key, detector in dg.items()})\n",
```
nvaytet, line 55:
```suggestion
    "dg = sc.DataGroup({key: detector[\"events\"][\"counter\", 0] for key, detector in dg.items()})\n",
```
to load only half the pixels and make the rendered page more lightweight.

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "    beam_direction=sc.vector([0, 0, 1]),\n",
+    "    wavelength=sc.linspace('wavelength', 4, 8, 20, unit='angstrom'),\n",
+    "    # To make it faster, don't compute the transmission fraction for every detector pixel, it's not necessary for the visualization.\n",
+    "    detector_position=dg['mantle'].coords['position']['strip', ::4]['wire', ::2].copy(),\n",
+    "    quadrature_kind='cheap',\n",
+    ")\n",
+    "\n",
+    "# The visualization expects the `position` coord to denote detector positions, and `wavelength` to be the last dimension.\n",
+    "transmission_fraction_mantle.coords['position'] = transmission_fraction_mantle.coords.pop('detector_position')\n",
+    "dream.instrument_view(transmission_fraction_mantle.transpose((*set(transmission_fraction_mantle.dims) - {'wavelength'}, 'wavelength')), dim='wavelength', camera=Camera(position=sc.vector((-2, 2, -2), unit='m'), look_at=sc.vector((1, -1/3, 1/2), unit='m')))"
```
nvaytet, line 118:
```suggestion
    "dream.instrument_view(transmission_fraction_mantle, dim='wavelength', pixel_size=20.0)"
```
Not sure setting the camera is so important, since we are also able to move the camera around in the online docs.
It adds a lot of noise to the notebook, which can scare away some users?

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,227 @@
...
```
nvaytet, line 166:
```suggestion
    "dream.instrument_view(transmission_fraction_mantle, dim='wavelength', pixel_size=20.0)"
```

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "    rod_shape,\n",
+    "    vanadium,\n",
+    "    beam_direction=sc.vector([0, 0, 1]),\n",
+    "    wavelength=sc.linspace('wavelength', 4, 8, 20, unit='angstrom'),\n",
+    "    detector_position=dg['endcap_backward'].coords['position']['strip', 0].copy(),\n",
+    "    quadrature_kind='cheap',\n",
+    ")\n",
+    "\n",
+    "transmission_fraction_endcap.coords['position'] = transmission_fraction_endcap.coords.pop('detector_position')\n",
+    "dream.instrument_view(transmission_fraction_endcap.transpose((*set(transmission_fraction_endcap.dims) - {'wavelength'}, 'wavelength')), dim='wavelength', camera=Camera(position=sc.vector((-1, 1, -3), unit='m'), look_at=sc.vector((0.3, -0.3, 1), unit='m')))"
```
nvaytet, line 194:
```suggestion
    "dream.instrument_view(transmission_fraction_endcap, dim='wavelength', pixel_size=20.0)"
```

nvaytet (commented):
So I finally had a look at making 2D figures, but I think what you have is fine.
I didn't realize initally that you also had a slider in the `wavelength` dimension.

Below are a few suggestions to make the notebook a little simpler and more lightweight.

jokasimr:
:+1: Good suggestions

nvaytet (approved):

PR #117, opened 2024-12-16
Add stap demo dream notebook

nvaytet:


nvaytet:
Not sure if this is still useful, after the time-of-flight workflow is incorporated in the Dream reduction workflow in #125

jl-wynen:
@nvaytet Should we close this PR?

PR #120, opened 2025-01-15
Bump to support python 3.13 and use scipp from nightly index.

MridulS:


requirements/nightly.txt
```diff
@@ -66,37 +74,75 @@ mpltoolbox==24.5.1
...
 pyyaml==6.0.2
     # via dask
+requests==2.32.3
+    # via pooch
 sciline @ git+https://github.com/scipp/sciline@main
     # via
     #   -r nightly.in
     #   essreduce
-scipp @ https://github.com/scipp/scipp/releases/download/nightly/scipp-nightly-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
+scipp==24.11.3.dev27+g47efd60e3
```
MridulS, line 145: This either needs to be fixed while building or uploading wheels to the index. It should ideally be a stable version number like `scipp==24.11.3.dev0` or the nightly requirements shouldn't be locked with a pinned requirements file.

PR #121, opened 2025-01-22
Select run norm via override_input

jl-wynen:
Requires https://github.com/scipp/sciline/pull/192

Uses `override_input` to select a normalisation. The API is no better than before. But we can lift the selection out of the `DreamGeant4Workflow` constructor and do it in a separate step, a bit like the suggestion in https://github.com/scipp/essdiffraction/pulls?q=is:pr%20is:closed#discussion_r1774768818


jl-wynen:
Closing as there seems to be no interest in pursuing this further.

PR #127, opened 2025-01-30
Link pooch to scicat datasets

jl-wynen:


src/ess/dream/data.py
```diff
@@ -79,6 +82,9 @@ def simulated_diamond_sample() -> str:
 def simulated_vanadium_sample() -> str:
     """Path to a GEANT4 CSV file for a vanadium sample.
 
+    SciCat:
```
nvaytet, line 85: Missing scicat identifier?

src/ess/dream/data.py
```diff
@@ -49,6 +49,9 @@ def get_path(name: str, unzip: bool = False) -> str:
 def simulated_diamond_sample() -> str:
     """Path to a GEANT4 CSV file for a diamond sample.
 
+    SciCat: 20.500.12269/d17b7169-b963-4b23-9455-d7332b992184
```
nvaytet, line 52: Can we just put a single url that points to the right place?
At the moment, it's not obvious how to even reach the file from that number.

You have to append this to the url below, but then replace the `/` in the middle with `%2F`.
I can't find the right combination of search fields from the scicat home page to reach the dataset (test search, keyword, or even the big search at the top of the page)

jl-wynen: > You have to append this to the url below, but then replace the / in the middle with %2F.

This is why I didn't want to show a url but the PID. But I can put both.

> I can't find the right combination of search fields from the scicat home page to reach the dataset (test search, keyword, or even the big search at the top of the page)

I can't seem to find one either. Searching by PID works, but everything else doesn't. But that seems to be a bug in scicat. E.g., searching by owner group gives me all datasets, not just the ones for that group.

nvaytet: I feel the search on Scicat has been broken for a long time...

nvaytet (approved):

PR #131, opened 2025-02-20
Simplify generation of `event_time_zero` in assembly of detector data

nvaytet:


tests/dream/geant4_reduction_test.py
```diff
@@ -67,8 +68,8 @@
     CalibrationFilename: None,
     UncertaintyBroadcastMode: UncertaintyBroadcastMode.drop,
     DspacingBins: sc.linspace('dspacing', 0.0, 2.3434, 201, unit='angstrom'),
-    TofMask: lambda x: (x < sc.scalar(0.0, unit='us'))
-    | (x > sc.scalar(86e3, unit='us')),
+    TofMask: lambda x: (x < sc.scalar(0.0, unit='us').to(unit=elem_unit(x)))
+    | (x > sc.scalar(86e3, unit='us').to(unit=elem_unit(x))),
```
jl-wynen, line 72: This is not great for usability. But I don't see a good solution. Unless we decide to allow mixed units in operations.

nvaytet: I know, it's not the best. But it is made worse by being a `lambda`. Could be written better in a normal function.
Also, the `elem_unit` will go away soon.

jl-wynen: I am more concerned about how easy it is to write filters. For readability, you are riht that a regular function may help.

jl-wynen (approved):

PR #136, opened 2025-03-10
Use instrument configuration to find time-of-flight lookup table

### LINKED ISSUE #132: Replace the TimeOfFlightLookupTableFilename with the InstrumentConfiguration
In the latest version of the powder reduction, we have a `TimeOfFlightLookupTableFilename` parameter which is used to load a tof lookup table from file.

We are meant to have a library of tables to chose from, depending on the chopper config.
Instead of requiring the user to know the filename, we should just use the `InstrumentConfiguration` enum which would then find the corresponding file automatically.

The enum would just be a dropdown in the widgets, which makes it easy to switch between configurations.

nvaytet:
This is to improve the usability of the workflow: we now use the instrument configuration Enum to automatically select which time-of-flight lookup table to use.

This means that the user does not need to know the file name of the lookup table, just selects the configuration which is now displayed as dropdown options in the widgets.

Fixes #132


jl-wynen (approved):
Looks good!

Is there a way to find out from a nexus file which configuration was used?

nvaytet:
> Is there a way to find out from a nexus file which configuration was used?

Not yet, I don't think... would be useful

PR #137, opened 2025-03-10
docs: some fixes after comments from celine

jokasimr:


docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -105,14 +107,16 @@
...
     "    # To make it faster, don't compute the transmission fraction for every detector pixel, it's not necessary for the visualization.\n",
     "    detector_position=dg['mantle'].coords['position']['strip', ::4]['wire', ::2].copy(),\n",
     "    quadrature_kind='cheap',\n",
     ")\n",
     "\n",
-    "# The visualization expects the `position` coord to denote detector positions, and `wavelength` to be the last dimension.\n",
+    "# The visualization expects the `position` coord to denote detector positions,\n",
     "transmission_fraction_mantle.coords['position'] = transmission_fraction_mantle.coords.pop('detector_position')\n",
+    "# and it expects `wavelength` to be the last dimension\n",
+    "transmission_fraction_mantle = transmission_fraction_mantle.transpose((*(set(transmission_fraction_mantle.dims) - {'wavelength'}), 'wavelength'))\n",
```
nvaytet, line 119:
```suggestion
```

I don't think that's necessary, I think it just works because we have `dim='wavelength'` below.

docs/user-guide/dream/dream-visualize-absorption.ipynb
```diff
@@ -130,75 +134,30 @@
...
     "    vanadium,\n",
     "    beam_direction=sc.vector([0, 0, 1]),\n",
-    "    wavelength=sc.linspace('wavelength', 4, 8, 20, unit='angstrom'),\n",
+    "    wavelength=sc.geomspace('wavelength', 0.5, 4, 20, unit='angstrom'),\n",
     "    detector_position=dg['endcap_backward'].coords['position']['strip', 0].copy(),\n",
     "    quadrature_kind='cheap',\n",
     ")\n",
     "\n",
     "transmission_fraction_endcap.coords['position'] = transmission_fraction_endcap.coords.pop('detector_position')\n",
+    "transmission_fraction_endcap = transmission_fraction_endcap.transpose((*(set(transmission_fraction_endcap.dims) - {'wavelength'}), 'wavelength'))\n",
```
nvaytet, line 158:
```suggestion
```

docs/user-guide/dream/dream-visualize-absorption.ipynb
jokasimr: If we run `method="medium"` or `method="expensive"` instead of `method="cheap" the transmission fraction is evaluated piecewise and not for the entire detector at once. This makes the returned value have a different shape. The instrument view requires the `wavelength` dimension to be the last dimension of the data array.

jokasimr: Maybe we should make the instrument view automatically transpose the input data it receives if it has the wrong shape.

nvaytet: > The instrument view requires the wavelength dimension to be the last dimension of the data array.

Where does it fail? Is it when it tries to flatten the dimensions?

> Maybe we should make the instrument view automatically transpose the input data it receives if it has the wrong shape.

Yes, we should. The user should not have to worry about the order of dimensions.

jokasimr: > Where does it fail? Is it when it tries to flatten the dimensions?

Yes exactly.

```
File [~/work/essdiffraction/src/ess/dream/instrument_view.py:64](http://localhost:8889/lab/tree/docs/user-guide/dream/src/ess/dream/instrument_view.py#line=63), in _pre_process(da, dim)
     62 if dim is not None:
     63     dims.remove(dim)
---> 64 out = da.flatten(dims=dims, to="pixel")
     65 sel = sc.isfinite(out.coords["position"])
     66 return out[sel]

File [~/mambaforge/envs/essdiffraction/lib/python3.10/site-packages/scipp/core/shape.py:363](http://localhost:8889/home/johannes/mambaforge/envs/essdiffraction/lib/python3.10/site-packages/scipp/core/shape.py#line=362), in flatten(x, dims, to)
    357 if to is None:
    358     # Note that this is a result of the fact that we want to support
    359     # calling flatten without kwargs, and that in this case it semantically
    360     # makes more sense for the dims that we want to flatten to come first
    361     # in the argument list.
    362     raise ValueError("The final flattened dimension is required.")
--> 363 return _call_cpp_func(_cpp.flatten, x, dims, to)

File [~/mambaforge/envs/essdiffraction/lib/python3.10/site-packages/scipp/core/_cpp_wrapper_util.py:27](http://localhost:8889/home/johannes/mambaforge/envs/essdiffraction/lib/python3.10/site-packages/scipp/core/_cpp_wrapper_util.py#line=26), in call_func(func, *args, **kwargs)
     25     return data_group_nary(func, *args, **kwargs)
     26 if out is None:
---> 27     return func(*args, **kwargs)
     28 else:
     29     return func(*args, **kwargs, out=out)

DimensionError: Can only flatten a contiguous set of dimensions in the correct order
```

nvaytet: Will you add the automatic transpose here or in a different PR?

src/ess/dream/instrument_view.py
```diff
@@ -44,6 +44,14 @@ def instrument_view(
     """
     from plopp.widgets import Box
 
+    if dim and isinstance(data, sc.DataArray | sc.DataGroup) and dim != data.shape[-1]:
+        dims = (
+            *data.dims[: data.dims.index(dim)],
+            *data.dims[data.dims.index(dim) + 1 :],
+            dim,
+        )
+        data = data.transpose(dims)
```
nvaytet, line 53:
```suggestion
        data = data.transpose([d for d in data.dims if d !=dim] + [dim])
```
?

src/ess/dream/instrument_view.py
```diff
@@ -44,6 +44,9 @@ def instrument_view(
     """
     from plopp.widgets import Box
 
+    if dim and isinstance(data, sc.DataArray) and dim in data.dims[:-1]:
```
nvaytet, line 47: Instead of skipping the case with a `DataGroup`, can we make a new datagroup where we transpose the dims of each element, if it contains `dim` in its dims?

jokasimr: We can do that :+1:

nvaytet (approved):

PR #139, opened 2025-03-12
Support figure kwargs in FlatVoxelViewer

### LINKED ISSUE #138: Add option to resize plot of FlatVoxelViewer (tool for DREAM)
Depending on the choice to be displayed along the axes, it could to useful to be able to increase the size of the plot in order to distinguish smaller features.
The figure is resizable by dragging the bottom right corner. But changing what to plot along the axes rescales the figure.

nvaytet:
This enables e.g. `FlatVoxelViewer(dg, figsize=(10, 8), cmap='magma')`

Fixes #138

### DIFF
--- a/src/ess/dream/diagnostics.py
+++ b/src/ess/dream/diagnostics.py
@@ -5,6 +5,7 @@
 import math
 from collections.abc import Callable, Iterable, Mapping
 from functools import reduce
+from typing import Any
 
 import ipywidgets as ipw
 import numpy as np
@@ -28,7 +29,7 @@ class FlatVoxelViewer(ipw.VBox):
         data: Mapping[str, sc.DataArray],
         *,
         rasterized: bool = True,
-        **kwargs: object,
+        **kwargs: Any,
     ) -> None:
         """Create a new viewer.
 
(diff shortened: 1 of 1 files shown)


src/ess/dream/diagnostics.py
```diff
@@ -149,8 +155,7 @@ def _flat_voxel_figure(
     data: sc.DataArray,
     horizontal_dim: str,
     vertical_dim: str,
-    *,
-    rasterized: bool = True,
+    **kwargs: object,
```
jokasimr, line 158: Why `object` instead of `Any`?

src/ess/dream/diagnostics.py
```diff
@@ -182,7 +187,7 @@ def _flat_voxel_figure(
     h_labels = [str(value) for value in h_coord.values]
     v_labels = [str(value) for value in v_coord.values]
 
-    fig = flat.plot(rasterized=rasterized, cbar=True)
+    fig = flat.plot(**kwargs)
```
jokasimr, line 190: Didn't we loose the `cbar=True` argument now?

nvaytet: Yes but I don't think it's needed? It works without it :shrug:

src/ess/dream/diagnostics.py
nvaytet: Don't know, it's `object` elsewhere in the file so I just used that. I can change to `dict[str, Any]` if you like?

jokasimr: I think that is better. Not sure if it actually makes a difference in this case, but if we pass the `kwargs` to a function that doesn't take `object` arguments the type checker might be smart enough to yell at us.

src/ess/dream/diagnostics.py
jokasimr: If you're confident it makes no difference then that's good enough for me.

jokasimr (approved):

src/ess/dream/diagnostics.py
celinedurniak: I am also fine with the change

celinedurniak (approved):

PR #140, opened 2025-03-18
Better landing page for docs

jl-wynen:
This copies over the landing page of ESSsans. It also includes some fixes / improvements. See individual comments.


docs/index.md
```diff
@@ -1,10 +1,63 @@
...
+   <style>
+    .transparent {display: none; visibility: hidden;}
+    .transparent + a.headerlink {display: none; visibility: hidden;}
+   </style>
+```
+
+```{role} transparent
+```
+
+# {transparent}`ESSdiffraction`
```
jl-wynen, line 24: This creates an invisible top level heading. It is invisible because the logo acts as the heading. But it is still good to have so that the `##` headings below are actually second level and don't get promoted to `#` heading.

I used the same implementation in Scitacean and it works well.

docs/index.md
```diff
@@ -1,10 +1,63 @@
...
+
+<span style="font-size:1.2em;font-style:italic;color:var(--pst-color-text-muted)">
```
jl-wynen, line 26: I changed this to use a variable for the color so that it adapts to the dark theme.

SimonHeybrock (approved):

PR #144, opened 2025-04-01
Subtract sample run by empty can

### LINKED ISSUE #141: Subtract empty can measurement in DREAM workflow
We have the data but we don't use it in the workflow. I seem to remember that there was something wrong with the data or something missing. But I don't remember what. Maybe we can just use it.

jl-wynen:
Fixes #141

This results in a lot of negative counts. Checking with IDS...

This still needs a way to disable background subtraction.

### DIFF
--- a/docs/user-guide/dream/dream-powder-reduction.ipynb
+++ b/docs/user-guide/dream/dream-powder-reduction.ipynb
@@ -56,7 +56,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "workflow = dream.DreamGeant4Workflow(run_norm=powder.RunNormalization.monitor_histogram)"
+    "workflow = dream.DreamGeant4Workflow(\n",
+    "    run_norm=powder.RunNormalization.monitor_histogram,\n",
+    "    subtract_empty_can=True,\n",
+    ")"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


jl-wynen:
I got a new vanadium simulation that doesn't result in negative counts. I am going to switch to that after #145 is merged to avoid conflicts.

jl-wynen:
This is now based on #160 and needs to wait until that is merged.

jl-wynen:
Rebased onto main

src/ess/powder/correction.py
```diff
@@ -335,6 +338,22 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
     return out
 
 
+def subtract_background(
+    data: FocussedDataDspacing[SampleRun],
+    background: FocussedDataDspacing[BackgroundRun],
+) -> BackgroundSubtractedData[SampleRun]:
+    return BackgroundSubtractedData[SampleRun](data.bins.concatenate(-background))
```
SimonHeybrock, line 345: Will we need the background subtraction to work also for histogrammed data?

jl-wynen: I don't know. For now, I have not seen histogrammed data at this point in the workflow.

docs/user-guide/dream/dream-powder-reduction.ipynb
```diff
@@ -94,7 +94,10 @@
     "workflow[DspacingBins] = sc.linspace(\"dspacing\", 0.3, 2.3434, 201, unit=\"angstrom\")\n",
     "\n",
     "# Do not mask any pixels / voxels:\n",
-    "workflow = powder.with_pixel_mask_filenames(workflow, [])"
+    "workflow = powder.with_pixel_mask_filenames(workflow, [])\n",
+    "\n",
+    "# Enable subtraction by an empty can measurement:\n",
+    "powder.correction.add_empty_can_subtraction(workflow)"
```
jl-wynen, line 100: I don't like this mechanism. I think a dedicated class for powder workflows would help here.

SimonHeybrock (approved):

PR #145, opened 2025-04-02
Split DREAM user guide into basic and advanced notebooks

jl-wynen:
This is an attempt to make the user guide more approachable. The basic notebook shows the simplest code to reduce some powder data. The advanced notebook shows how the reduction can be customised.

We may want to use different default values in the basic notebook eventually. In particular, we should figure out which normalisation will be the default.

Reviewer: please look at the rendered docs. In particular, take a look at the plots for integrated monitor norm and stacked detectors. I am not completely sure they are correct.

This supersedes #126. I think it is not feasible to have a separate notebook for each set of workflow params because that would lead to an exponential grows in notebooks. So in contrast to #126, this PR shows the individual customisation points in isolation.


nvaytet (commented):
Nice idea with the thumbnails. Two things:

- The basic thumbnails are displaying empty axes
- In Plopp, I saved a figures to the `_static` folder when running the notebooks, and then used those as thumbnails instead of uploading the svgs to the git repo. This is nice as we don't need to update the images when the workflow updates, but it also means having a hidden cell at the bottom of the notebook, which is annoying if the users download the notebook.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,480 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "14",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "result = workflow.compute(IofTof)\n",
+    "result.hist().plot()"
```
nvaytet, line 206: This plot is empty in the generated docs.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,480 @@
...
+   "source": [
+    "result"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "31",
+   "metadata": {},
+   "source": [
+    "We can plot the detectors individually with the help of `sc.collapse`:"
```
nvaytet, line 424: Instead of stacking the detectors, could we simply put them in a datagroup?
Then we don't have to remember that detector number 0 is mantle?

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,480 @@
...
+   "execution_count": null,
+   "id": "32",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "split = sc.DataGroup({\n",
+    "    da.coords['detector'].value: da\n",
+    "    for da in sc.collapse(result, keep='tof').values()\n",
+    "})\n",
+    "split.hist().plot()"
```
nvaytet, line 438: Some of the results look weird, with broken lines (maybe NaNs in the data?)

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: Not completely empty for me but definitely broken:
integrated

This seems to be a problem with the integrated monitor norm. I don't know what. But iit should not have been caused by this PR.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: We could. I thought it would be easier to process them all at once when they are stacked like this. Note that I added a 'detector' coord that contains the name.

But I can change this if you prefer a data group.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: > (maybe NaNs in the data?)

I think so.

But this is not a problem with the docs, AFAIK. It may be caused by the normalisation with slices the data based on the wavelength range of the monitor.

jl-wynen:
>     * The basic thumbnails are displaying empty axes

Fixed.

>     * In Plopp, I saved a figures to the `_static` folder when running the notebooks, and then used those as thumbnails instead of uploading the svgs to the git repo. This is nice as we don't need to update the images when the workflow updates, but it also means having a hidden cell at the bottom of the notebook, which is annoying if the users download the notebook.

Could do. I wouldn't want to do it with the basic workflow though, because that is supposed to be immediately usable when downloaded. So we would have two separate solutions for the different notebooks.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: The monitor contains NaNs. So some detector events also become NaN when normalising.

I could hide this from the basic docs by using the proton charge normalisation. Should I?

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
nvaytet: We should really have a test in place that catches this...

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
nvaytet: Personally, I would go with a DataGroup.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: The plot looks fine now

src/ess/powder/calibration.py
```diff
@@ -94,19 +94,19 @@ def to_cif_format(self) -> sc.DataArray:
...
 def assemble_output_calibration(
-    data: DspacingData[SampleRun],
+    beamline: CalibratedBeamline[SampleRun],
+    coord_transform_graph: ElasticCoordTransformGraph,
 ) -> OutputCalibrationData:
-    """Construct output calibration data from average pixel positions."""
-    # Use nanmean because pixels without events have position=NaN.
-    average_l = sc.nanmean(data.coords["Ltotal"])
-    average_two_theta = sc.nanmean(data.coords["two_theta"])
+    """Construct output calibration data from fake pixel positions."""
```
nvaytet, line 100: Revert to old implementation.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,572 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "10",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "workflow = dream.DreamGeant4Workflow(run_norm=powder.RunNormalization.monitor_integrated)"
```
nvaytet, line 152: Put in the same cell as params for easy copy/paste

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,572 @@
...
+    "workflow[dream.InstrumentConfiguration] = dream.InstrumentConfiguration.high_flux\n",
+    "# Select a detector bank:\n",
+    "workflow[NeXusDetectorName] = \"mantle\"\n",
+    "# We drop uncertainties where they would otherwise lead to correlations:\n",
+    "workflow[UncertaintyBroadcastMode] = UncertaintyBroadcastMode.drop\n",
+    "# Edges for binning in d-spacing:\n",
+    "workflow[DspacingBins] = sc.linspace(\"dspacing\", 0.3, 2.3434, 201, unit=\"angstrom\")\n",
+    "\n",
+    "# Do not mask any pixels / voxels:\n",
+    "workflow = powder.with_pixel_mask_filenames(workflow, [])"
```
nvaytet, line 189: Show all workflow graphs?

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -0,0 +1,620 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "28",
+   "metadata": {},
+   "source": [
+    "At this point, the workflow is incomplete because it is missing the `NexisDetectorName` parameter.\n",
```
nvaytet, line 395:
```suggestion
    "At this point, the workflow is incomplete because it is missing the `NexusDetectorName` parameter.\n",
```

nvaytet (approved):

nvaytet:
Interesting this line on the left edge of the endcap_forward (orange):
Screenshot_20250507_104617

It seems it is a line because other banks show a line there, but it gets boosted by the low counts in the vanadium...

jl-wynen:
> It seems it is a line because other banks show a line there, but it gets boosted by the low counts in the vanadium...

Agree. If you look at the unnormalised data, it has a line there but a much weaker one.

PR #148, opened 2025-04-04
Mask zero values in the monitor

jl-wynen:
~This should make the norm more robust. But I don't know if this is really the best way. Should we remove or mask detector elements where the monitor is NaN?~
This fixes https://github.com/scipp/essdiffraction/pull/145#discussion_r2026991380

**Edit:** We used to replace zero counts in the monitor with NaNs. Instead of doing that, we mask the values.


SimonHeybrock:
> I don't know if this is really the best way. Should we remove or mask detector elements where the monitor is NaN? 

@celinedurniak Do you have an opinion?

celinedurniak:
For which options when normalizing by monitor does this issue occur?

nvaytet:
The NaNs are introduced by the computation of time-of-flight.
If we are looking for an `event_time_offset` or `Ltotal` outside the lookup table, the tof will be given a NaN value.

You can see this in the resulting monitor data/plot (at the low and high tof ends):
Screenshot_20250407_103858
Screenshot_20250407_104221

However, I'm still not sure why we are getting NaNs in the result.
This is a histogrammed monitor, correct?
If we are computing the `ResampledTofData`, it should be rebinning to exclude all bins that have nans as coords: https://github.com/scipp/essreduce/blob/main/src/ess/reduce/time_of_flight/eto_to_tof.py#L628

So we should be excluding all nans?


**Edit:** I believe the NaNs actually come from [here](https://github.com/scipp/essdiffraction/blob/main/src/ess/powder/conversion.py#L321) where we are setting all zeros in the monitor to NaN (presumably to avoid divisions by zero?).
Is this the issue and maybe this should be changed instead?

jl-wynen:
I don't think setting it to NaN is a good idea here. It looks like the tof range is too wide and it includes bins where we cannot measure anything. So I think we should either mask them or adjust the binning.

nvaytet:
> So I think we should either mask them or adjust the binning.

Does masking help? We would divide the signal by masked bins. We would still end up with infinite values, that are masked. I guess that is better, but is it what we want?

nvaytet:
Let's try to go ahead with masking?

jl-wynen:
What do the plots look like with this?

nvaytet:
> What do the plots look like with this?

They look good.... at least on my workstation.

**Monitor plot**
Figure 1 (12)

**RunNormalization.monitor_integrated**
Figure 1 (13)


Now that I try at home on my laptop, the `RunNormalization.monitor_histogram` results are masked everywhere...
Figure 1 (14)

Note sure what's up... some package has a different version somewhere.

nvaytet:
Ok I updated everything and now it looks good also for `monitor_histogram`
Figure 1 (15)

PR #152, opened 2025-04-11
Update lookup table to 2D version

nvaytet:
Builds will fail until changes in https://github.com/scipp/essreduce/pull/230 are released.


nvaytet:
Superseded by #160

PR #155, opened 2025-04-22
Fix reduction of all DREAM banks

jl-wynen:
Note that this is a PR to merge into #145.

I think this PR fixes the issues with reducing multiple banks together.

I had to change the calibration provider to not use actual detector positions because it only has access to one bank at a time. But to merge all banks into a single I(t) array, we need a single calibration for all banks. Since the calibration is fake anyway, the new approach should be fine.

### DIFF
--- a/docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
+++ b/docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
@@ -24,13 +24,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from collections.abc import Iterable\n",
-    "\n",
     "import scipp as sc\n",
     "from ess import dream, powder\n",
     "import ess.dream.data  # noqa: F401\n",
     "from ess.powder.types import *\n",
-    "import pandas as pd"
+    "import pandas as pd\n",
+    "import plopp as pp"
    ]
   },
   {
@@ -471,7 +470,9 @@
     "### 2D intensity vs. d-spacing and 2θ\n",
     "\n",
     "Just like before, we map the workflow over the detector names.\n",
-    "But here we also assign different $2\\theta$ bins for each detector bank."
+    "But here we also assign different $2\\theta$ bins for each detector bank.\n",
+    "Note that the $2\\theta$ ranges don’t overlap.\n",
+    "This will simplify later steps."
    ]
   },
   {
@@ -486,10 +487,12 @@
     "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=0.77, stop=2.36, num=201),\n",
     "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=0.24, stop=0.71, num=101),\n",
     "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=2.42, stop=2.91, num=151),\n",
-    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=2.94, stop=3.11, num=51),\n",
+    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=2.91, stop=3.11, num=51),\n",
     "]\n",
     "parameter_table = pd.DataFrame(\n",
-    "    {NeXusDetectorName: detector_names, TwoThetaBins: two_theta_bins},\n",
+    "    {NeXusDetectorName: detector_names,\n",
+    "     TwoThetaBins: two_theta_bins,\n",
+    "     },\n",
     "    index=detector_names\n",
     ").rename_axis(index='detector')\n",
     "\n",
@@ -513,7 +516,7 @@
    "id": "37",
    "metadata": {},
    "source": [
-    "We can 'sum' over $2\\theta$ by using `concat` and then sum over all detectors to obtain a 1D curve:"
+    "We can 'sum' over $2\\theta$ by using `concat` to obtain a 1D curve:"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -453,7 +461,134 @@
...
+   "id": "36",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "detector_names = [\"mantle\", \"endcap_forward\", \"endcap_backward\", \"high_resolution\"]\n",
+    "two_theta_bins = [\n",
+    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=0.77, stop=2.36, num=201),\n",
+    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=0.24, stop=0.71, num=101),\n",
+    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=2.42, stop=2.91, num=151),\n",
+    "    sc.linspace(dim=\"two_theta\", unit=\"rad\", start=2.94, stop=3.11, num=51),\n",
```
nvaytet, line 498: I find it annoying that we have to know all these values.
We figured it out by plotting the data for each one, but it's difficult for a user (instrument scientist) to know theme without plotting first...

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -453,7 +461,134 @@
...
+   "id": "40",
+   "metadata": {},
+   "source": [
+    "But we can also preserve the $2\\theta$ dimension to produce a 2D plot.\n",
+    "However, the detector banks have different $2\\theta$ ranges.\n",
+    "So we need to merge them into an appropriate array first.\n",
+    "We could concatenate the event lists for all banks directly and then bin into new $2\\theta$ and $d$-spacing bins.\n",
+    "\n",
+    "But we can also preserve the $2\\theta$ binning for each bank to respect differences in resolution.\n",
+    "The following cell does so by filling the gaps between banks with empty event lists.\n",
```
nvaytet, line 549: It seems very cumbersome to have to manually add empty event lists.
I know you didn't like to wast memory by making common 2theta bins for all, but it wasn't that much memory (the dimensions are not large).
I feel it would be more convenient to work with?

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
```diff
@@ -453,7 +461,134 @@
...
-    "result.hist(tof=300).plot()"
+    "histograms = []\n",
+    "for events in result.values():\n",
+    "    hist = events.hist()\n",
+    "    # Apply the mask\n",
+    "    if (mask := hist.masks.pop('zero_vanadium', None)) is not None:\n",
+    "        hist.data = sc.where(mask, sc.scalar(0.0), hist.data)\n",
+    "    histograms.append(hist)\n",
+    "\n",
+    "sc.concat(histograms, dim='detector').sum('detector').plot()"
```
nvaytet, line 472: I can't figure out in my head if this is a valid thing to do.
To me, the plot below where we first convert to d-spacing and 2theta, and then sum over 2theta makes sense.
This one does something different (the plots look different at least).

I guess I'm not sure why they are different...

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: True. Should we compute them based on geometry and store them in the package? To do this precisely, we would need the voxel positions and sized. The simulation only has event (sub-voxel) positions.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: It may be. But doing so is also a bit complicated when we want different resolutions for the banks.
Let me check with Céline to see what we actually need.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
jl-wynen: They are normalised differently. This one here computes an I(d) curve for both sample and vanadium and divides the two. The other case computes an I(2theta, d) curve for sample and vanadium, divides them, and only then removes the 2theta dimension.

nvaytet: From today's discussion, it sounds like we need to compute 2d maps (2theta - dspacing) for sample and vanadium, divide them, and then sum along 2theta?

jl-wynen: I took that away from it, too. But I don't like introducing an arbitrary binning in 2theta in an intermediate step. (When we produce 1D outputs.)

nvaytet: Instead of having something arbitrary, you could use the mean pixel size and mean distance to compute a theta resolution?

src/ess/powder/calibration.py
```diff
@@ -94,18 +94,19 @@ def to_cif_format(self) -> sc.DataArray:
 
 
 def assemble_output_calibration(
-    beamline: CalibratedBeamline[SampleRun],
+    data: DspacingData[SampleRun],
```
nvaytet, line 97: This doesn't seem quite right.

Inside this function, you are only using static coordinates `Ltotal` and `two_theta`.
If you use `DspacingData` instead of `CalibratedBeamline`, it means that when you are streaming live data, you would recompute these positions every time some new events come in,

We want to avoid this by using `CalibratedBeamline` instead.

jl-wynen: I did this to support processing multiple banks. I wanted to have a single calibration table which requires access to all banks at the same time. But `CalibratedBeamline` is mapped over banks. The earliest value that is merged is `DspacingData`.

However, I am going to change it back and not merge the banks and instead process them separately. Then `assemble_output_calibration` can be reverted.

docs/user-guide/dream/dream-advanced-powder-reduction.ipynb
nvaytet: I would have hoped we could just use a constant bin width for everyone, and let the binning/histogramming figure out the bounds?

jl-wynen: For the record: We tried to use a common number of bins or a common bin width instead of mapping over edges. This leads to different binning in sample and vanadium data which means that `_normalize_by_vanadium` needs to be adjusted. Doing so is non trivial because it has to deal with 2D data. Since in practice, we expect people to use specific, known-to-be-good bin edges that are determined by scientists, we stick with the current implementation of mapping over edges.

src/ess/powder/calibration.py
jl-wynen: Reverted

jl-wynen:
@nvaytet Did I address everything in this PR?

nvaytet (approved):

PR #160, opened 2025-05-05
Use generic tof workflow

nvaytet:
Needs https://github.com/scipp/essreduce/pull/233

### DIFF
--- a/docs/user-guide/sns-instruments/POWGEN_data_reduction.ipynb
+++ b/docs/user-guide/sns-instruments/POWGEN_data_reduction.ipynb
@@ -239,7 +239,7 @@
    "source": [
     "results = workflow.compute(\n",
     "    (\n",
-    "        TofData[SampleRun],\n",
+    "        DetectorTofData[SampleRun],\n",
     "        MaskedData[SampleRun],\n",
     "        FilteredData[SampleRun],\n",
     "        FilteredData[VanadiumRun],\n",
@@ -254,7 +254,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "results[TofData[SampleRun]]"
+    "results[DetectorTofData[SampleRun]]"
    ]
   },
   {


src/ess/dream/data.py
```diff
@@ -27,7 +27,7 @@ def _make_pooch():
             "DREAM_simple_pwd_workflow/Cave_TOF_Monitor_diam_in_can.dat": "md5:ef24f4a4186c628574046e6629e31611",  # noqa: E501
             "DREAM_simple_pwd_workflow/Cave_TOF_Monitor_van_can.dat": "md5:e63456c347fb36a362a0b5ae2556b3cf",  # noqa: E501
             "DREAM_simple_pwd_workflow/Cave_TOF_Monitor_vana_inc_coh.dat": "md5:701d66792f20eb283a4ce76bae0c8f8f",  # noqa: E501
-            "DREAM-high-flux-tof-lookup-table.h5": "md5:404145a970ed1188e524cba10194610e",  # noqa: E501
+            "DREAM-high-flux-tof-lookup-table.h5": "md5:1b95a359fa7b0d8b4277806ece9bf279",  # noqa: E501
```
jl-wynen, line 30: Is this the 2d table?

src/ess/dream/workflow.py
```diff
@@ -85,7 +85,7 @@ def default_parameters() -> dict:
 def _collect_reducer_software() -> ReducerSoftwares:
     return ReducerSoftwares(
         [
-            Software.from_package_metadata('essdiffraction'),
+            # Software.from_package_metadata('essdiffraction'),
```
jl-wynen, line 88: Why?

jl-wynen (commented):
Looks good, just some minor comments.

src/ess/dream/workflow.py
nvaytet: Left over from local development.
It's an issue for me because of the way I locally 'installed' our packages (using symbolic links instead of a `pip install`).
I'll change it back.

src/ess/dream/data.py
nvaytet: Yes, with additional coords (I think I updated the notebook that generates it?)

nvaytet:
The issue with the docs build is apparently when I insert [this provider](https://github.com/scipp/essdiffraction/pull/160/files#diff-43b5bede067c6b8edaaaf39218b007b2f3c0a2a8bd0eb84c4bde6c4e465f9e2cR121)

Without it, no error in the docs.
I don't understand what is going on. The `get_possible_outputs` should be sorting a tuple with strings, there shouldn't be an issue.
If I print the content of `get_possible_outputs`, before it tries to sort it, I get a long list of types, but there is no `_GenericAlias` anywhere in that list... :thinking:

SimonHeybrock:
> The issue with the docs build is apparently when I insert [this provider](https://github.com/scipp/essdiffraction/pull/160/files#diff-43b5bede067c6b8edaaaf39218b007b2f3c0a2a8bd0eb84c4bde6c4e465f9e2cR121)
> 
> Without it, no error in the docs. I don't understand what is going on. The `get_possible_outputs` should be sorting a tuple with strings, there shouldn't be an issue. If I print the content of `get_possible_outputs`, before it tries to sort it, I get a long list of types, but there is no `_GenericAlias` anywhere in that list... 🤔

No generic type?

src/ess/dream/workflow.py
```diff
@@ -92,13 +98,27 @@ def _collect_reducer_software() -> ReducerSoftwares:
...
+def convert_dream_monitor_to_wavelength(
+    monitor: ResampledMonitorTofData[RunType, MonitorType],
+) -> WavelengthMonitor[RunType, MonitorType]:
+    """
+    We know that DREAM monitors are recording in histogram mode, so we need to use the
+    resampled monitor data to avoid having nans in the time-of-flight coordinates.
+
+    This provider should be inserted in the Dream workflow below.
+    """
+    return convert_monitor_to_wavelength(monitor)
```
SimonHeybrock, line 110: Why this wrapper?

src/ess/powder/conversion.py
```diff
@@ -247,81 +231,14 @@ def convert_reduced_to_tof(
...
-    wf[time_of_flight.DistanceResolution] = distance_resolution
-    wf[time_of_flight.TimeResolution] = time_resolution
-    wf[time_of_flight.LookupTableRelativeErrorThreshold] = error_threshold
-    return wf.compute(time_of_flight.TimeOfFlightLookupTable)
-
-
 def load_tof_lookup_table(
     filename: TimeOfFlightLookupTableFilename,
 ) -> TimeOfFlightLookupTable:
     return TimeOfFlightLookupTable(sc.io.load_hdf5(filename))
```
SimonHeybrock, line 237: Is this still used? If so, why is there no such provider in essreduce?

src/ess/powder/types.py
```diff
@@ -213,18 +213,6 @@ class RawDataAndMetadata(sciline.Scope[RunType, sc.DataGroup], sc.DataGroup):
...
-
-class MonitorLtotal(sciline.Scope[RunType, MonitorType, sc.Variable], sc.Variable):
-    """Total path length of neutrons from source to monitor."""
-
-
-class TofMonitorData(sciline.Scope[RunType, MonitorType, sc.DataArray], sc.DataArray):
-    """Monitor data with time-of-flight coordinate."""
-
-
 TimeOfFlightLookupTableFilename = NewType("TimeOfFlightLookupTableFilename", str)
```
SimonHeybrock, line 216: Is this not generic across instruments? Why not define in essreduce?

nvaytet: Yes, let's do that

nvaytet: Since you can either load a lookup table from file, or build it from a simulation.

I'm assuming that the default should be that you want to load the table from a file (this is what most people will want to do).
What would be a convenient way to switch between this mode, and building from simulation?

At the moment, i'm having to know which provider to import and then insert in the workflow, as done in the tests here: https://github.com/scipp/essdiffraction/pull/160/files#diff-397d502aff143b028a421a4e43ded449baae3d8b2ae6711bee74734d7aa1c9e5R146

Is this ok for now? It's not many lines of code, but knowing the import path is annoying. Should I add that particular provider to the `__init__` of the `time_of_flight` submodule to make it more accessible?

SimonHeybrock: When computing the LUT, don't you actually need a whole bunch of extra bits in the pipeline, such as loading choppers?

nvaytet: Hmm yes. It seems that in the test here I've stopped  the workflow at the `SimulationResults`, which is computed once, without reading choppers from the files.

However, in the TBL workflow in `essimaging`, I'm using the choppers from the file in a provider to make the `SimulationResults`: https://github.com/scipp/essimaging/blob/main/src/ess/tbl/conversion.py#L66
Looking at that provider, I'm thinking I should probably also read the `source_position` from the nexus file.

So we would need to insert at least these two providers. Should we make 2 different workflows? One that loads lookup table from file, and one that builds it on-the-fly?

nvaytet: That would probably be convenient. However, does this mean I now have 6 cases instead of the 3 [here](https://github.com/scipp/essdiffraction/blob/main/src/ess/dream/workflow.py#L112-L135)?
```Py
@register_workflow
def DreamGeant4MonitorHistogramWorkflow() -> sciline.Pipeline:
    """
    Workflow with default parameters for the Dream Geant4 simulation, using a
    histogrammed monitor for the normalization.
    """
    return DreamGeant4Workflow(run_norm=RunNormalization.monitor_histogram)


@register_workflow
def DreamGeant4MonitorIntegratedWorkflow() -> sciline.Pipeline:
    """
    Workflow with default parameters for the Dream Geant4 simulation, using
    integrated counts of the monitor for the normalization.
    """
    return DreamGeant4Workflow(run_norm=RunNormalization.monitor_integrated)


@register_workflow
def DreamGeant4ProtonChargeWorkflow() -> sciline.Pipeline:
    """
    Workflow with default parameters for the Dream Geant4 simulation, using
    proton charge for the normalization.
    """
    return DreamGeant4Workflow(run_norm=RunNormalization.proton_charge)
```

Won't this grow exponentially?
I guess this is mostly for the widgets, it doesn't really affect the rest if I add an argument to the underlying [DreamGeant4Workflow](https://github.com/scipp/essdiffraction/blob/main/src/ess/dream/workflow.py#L95)?

SimonHeybrock: You are right that some things are a bit too inflexible right now. Maybe we need to support workflow (creation) parameters in the widgets?

src/ess/dream/workflow.py
SimonHeybrock: Ignore this, the docstring explains.

nvaytet (commented):
Looks good.
Again, can't approve.
Note that dependencies will need updating again after we release essreduce==25.05.1.

SimonHeybrock (approved):

PR #170, opened 2025-05-21
Move empty can subtraction to end

jl-wynen:
This should make working with live data more efficient.

### DIFF
--- a/src/ess/powder/correction.py
+++ b/src/ess/powder/correction.py
@@ -13,10 +13,10 @@ from ._util import event_or_outer_coord
 from .types import (
     AccumulatedProtonCharge,
     BackgroundRun,
-    BackgroundSubtractedData,
-    BackgroundSubtractedDataTwoTheta,
     CaveMonitor,
     DataWithScatteringCoordinates,
+    EmptyCanSubtractedIofDspacing,
+    EmptyCanSubtractedIofDspacingTwoTheta,
     FocussedDataDspacing,
     FocussedDataDspacingTwoTheta,
     IofDspacing,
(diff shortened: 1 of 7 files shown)


src/ess/powder/correction.py
```diff
@@ -424,19 +339,17 @@ def _shallow_copy(da: sc.DataArray) -> sc.DataArray:
 
 
 def subtract_background(
-    data: FocussedDataDspacing[SampleRun],
+    data: IofDspacing,
     background: FocussedDataDspacing[BackgroundRun],
```
SimonHeybrock, line 343: Shoudn't this be `IofDspacing[BackgroundRun]`? In other words, don' we need to parametrize `IofDspacing`?

jl-wynen: Done. But this required changing `IofDspacing` to a generic type. So this is a breaking change to the workflow which needs to request `IofDspacing[SampleRun]` now. I had coded `SampleRun` in some places to avoid errors from using the same type twice as an argument. E.g.,
```python
def subtract_background(
    data: IofDspacing[RunType],
    background: IofDspacing[BackgroundRun],
) -> EmptyCanSubtractedIofDspacing[RunType]:
    ...
```
Would result in an instantiation that uses `IofDspacing[BackgroundRun]` for both arguments and Sciline doesn't like that.

SimonHeybrock: Given that tests passed with this "bug" (before your latest changes): Is there a test missing that actually checked the result of the background subtraction?

jl-wynen: We don't have any regression tests for the final result. But they would have been broken in this case anyway.
I don't know how we can test that the background subtraction is applied correctly. Unfortunately, the data are dimensionless at this point. Do you have an idea?

SimonHeybrock: Check that the workflow gives `(A - B)/C`, locally?

jl-wynen: See new commit

tests/dream/geant4_reduction_test.py
```diff
@@ -187,11 +183,36 @@ def test_pipeline_can_compute_dspacing_result_with_integrated_monitor_norm(
...
+    result = results[EmptyCanSubtractedIofDspacing[SampleRun]]
+
+    sample = results[FocussedDataDspacing[SampleRun]]
+    empty_can = results[FocussedDataDspacing[BackgroundRun]]
+    vanadium = results[FocussedDataDspacing[VanadiumRun]]
+
+    expected = powder.correction.normalize_by_vanadium_dspacing(
+        sample.bins.concatenate(-empty_can), vanadium, UncertaintyBroadcastMode.drop
+    )
+    sc.testing.assert_allclose(result, expected)
```
SimonHeybrock, line 213: Ok, but I thought we wanted to avoid always running a full workflow in every "unit" test, due to growing run times... 😬

SimonHeybrock (approved):

tests/dream/geant4_reduction_test.py
jl-wynen: Good point. This was easy enough to change.

PR #177, opened 2025-06-02
Update lookup table generating tool using tof.

YooSunYoung:
CL needed this and the current notebook didn't work with the latest packages anymore.

I just updated it to be same as the one I made for easyimaging.


YooSunYoung:
> Note that if we decided to go ahead with https://github.com/scipp/essreduce/pull/253, then this notebook would change again. So we may want to wait until that has been released? 

We'll come back to it after https://github.com/scipp/essreduce/pull/253 is sorted.

PR #183, opened 2025-06-20
Add BC240 TofLUT.

SimonHeybrock:
This is needed for McStas simulation files shared by Celine.

### DIFF
--- a/src/ess/dream/data.py
+++ b/src/ess/dream/data.py
@@ -2,6 +2,8 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 """Data for tests and documentation with DREAM."""
 
+from typing import Literal
+
 _version = "2"
 
 __all__ = ["get_path"]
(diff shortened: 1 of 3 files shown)


src/ess/dream/data.py
```diff
@@ -278,5 +282,17 @@ def tof_lookup_table_high_flux() -> str:
```
SimonHeybrock, line 282: This docstring claims the old file was for 240 degrees already, but @celinedurniak mentioned on Slack it is 215. @nvaytet do you remember the history of this?

nvaytet: I think I it's simply that the old docstring is wrong

nvaytet:
Note that I believe it is not the plan for dream to switch between 215 and 240, but more to settle on which one is best and just have that as the 'high flux'.

But still useful to be able to switch between the two while we are still on the development stage

nvaytet (approved):

PR #185, opened 2025-06-27
Reorganise Dream reduction workflows

nvaytet:
After introducing the `GenericTofWorkflow`, the old `dream.io.nexus.LoadNeXusWorkflow` which reshaped detector banks was no longer used.

This PR removed that old nexus workflow, and re-implements bank reshaping in the new workflows.

We now have a `DreamGenericWorkflow` which is basically the `GenericTofWorkflow` with detector bank reshaping.
Then, the `DreamPowderWorkflow` and the `DreamGeant4Workflow` build on top of this workflow.

I did not have to touch the docs, so hopefully this means the API didn't change?


src/ess/dream/workflows/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+__all__ = [
+    'DreamGeant4MonitorHistogramWorkflow',
+    'DreamGeant4MonitorIntegratedWorkflow',
+    'DreamGeant4ProtonChargeWorkflow',
+    'DreamGeant4Workflow',
+    'DreamGenericWorkflow',
+    'DreamPowderWorkflow',
+]
```
SimonHeybrock, line 23: Not sure why we need this subfolder. Confusing why some workflows are in `dream`, others only in `dream.workflows`. Why not have everything in `dream`? We won't have 1000 workflows so I think that should be fine?

src/ess/dream/workflows/generic.py
```diff
@@ -0,0 +1,91 @@
...
+    return ReducerSoftware(
+        [
+            Software.from_package_metadata('essdiffraction'),
+            Software.from_package_metadata('scippneutron'),
+            Software.from_package_metadata('scipp'),
+        ]
+    )
+
+
+def DreamGenericWorkflow() -> sciline.Pipeline:
```
SimonHeybrock, line 74: DreamWorkflow?

src/ess/dream/workflows/generic.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Dream generic workflow with default parameters.
+    The workflow is based on the GenericTofWorkflow.
```
SimonHeybrock, line 77: Maybe explain what it can do, i.e., load and compute Tof?

tests/dream/io/geant4_test.py
```diff
@@ -176,18 +176,20 @@ def test_load_geant4_csv_sans_has_expected_coords(file):
...
+        wf.insert(provider)
+    wf[Filename[SampleRun]] = file_path
+    wf[NeXusDetectorName] = NeXusDetectorName("mantle")
+    wf[NeXusComponent[snx.NXsample, SampleRun]] = sc.DataGroup(
         position=sc.vector([0.0, 0.0, 0.0], unit="mm")
     )
-    pipeline[NeXusComponent[snx.NXsource, SampleRun]] = sc.DataGroup(
+    wf[NeXusComponent[snx.NXsource, SampleRun]] = sc.DataGroup(
         position=sc.vector([-3.478, 0.0, -76550], unit="mm")
     )
```
SimonHeybrock, line 190: Not sure why this is not using the Geant4 workflows, where these things are defined already?

src/ess/dream/workflows/generic.py
nvaytet: I thought having Generic in there made it a little more obvious that this should be used as a base for other workflows, but I guess in principle you can also use it on its own.
So I don't have a super strong opinion.

src/ess/dream/workflows/__init__.py
SimonHeybrock: My comment was about which you have in the respective `__init__.py` on the various levels. Also, we should avoid too much user-visible nesting, since it is just complicated and confusing.

SimonHeybrock (approved):

PR #186, opened 2025-07-01
Avoid masks on bin edges

nvaytet:
Check for bin edge coord when masking to avoid having masks on edges.

### DIFF
--- a/src/ess/powder/masking.py
+++ b/src/ess/powder/masking.py
@@ -63,7 +63,7 @@ def apply_masks(
             else:
                 coord = (
                     sc.midpoints(out.coords[dim])
-                    if out.coords.is_edges(dim)
+                    if out.coords.is_edges(dim, dim)
                     else out.coords[dim]
                 )
                 out.masks[dim] = mask(coord)


nvaytet:
It seems all the masks were being applies on events, so this was being tested?

It also seems that the masks like tof or wavelength masks are not applied to the monitors?
Screenshot_20250701_165327

It also doesn't appear to be hidden inside the [code that does the normalization](https://github.com/scipp/essdiffraction/blob/main/src/ess/powder/correction.py#L70)?

SimonHeybrock:
> It seems all the masks were being applies on events, so this was being tested?
> 
> It also seems that the masks like tof or wavelength masks are not applied to the monitors? Screenshot_20250701_165327
> 
> It also doesn't appear to be hidden inside the [code that does the normalization](https://github.com/scipp/essdiffraction/blob/main/src/ess/powder/correction.py#L70)?

See #166... but I think I tried to fix that in #176, maybe it should have been closed, unless there are more cases?

nvaytet:
Not sure I fully followed. In the normalization there, it seems to be relying on `nanmin/max` of coords, but it doesn't seem to be taking wavelength masks into account?

In the case of binned data, it's getting the hi/lo from [nanmin on the event coords](https://github.com/scipp/essdiffraction/pull/176/files#diff-fd0b851aa3bf0cd4e34ba9a12e34aca475ce399ced3b52d5044f2615b36ac3a5R112), not taking any masks into account?

PR #190, opened 2025-07-15
Update to use the new tof lookup table workflow

nvaytet:
This is a necessary update after https://github.com/scipp/essreduce/pull/253

Only the tests were updated, as the main mode of operation is to use a lookup table read from a file.


jl-wynen (commented):
The changes look good. Can you update the lower pin?

Also, did you check whether https://github.com/scipp/essdiffraction/blob/main/tools/dream-make-tof-lookup-table.ipynb still works?

docs/user-guide/dream/dream-make-tof-lookup-table.ipynb
```diff
@@ -54,15 +57,12 @@
...
-    "    time_of_flight.providers(), params=time_of_flight.default_parameters()\n",
-    ")\n",
+    "wf = time_of_flight.TofLookupTableWorkflow()\n",
     "\n",
     "wf[time_of_flight.LtotalRange] = sc.scalar(60.0, unit=\"m\"), sc.scalar(80.0, unit=\"m\")\n",
-    "wf[time_of_flight.SimulationResults] = time_of_flight.simulate_beamline(\n",
-    "    choppers=disk_choppers, neutrons=5_000_000, source_position=sc.vector([0, 0, 0], unit='m'),\n",
-    ")\n",
-    "\n",
+    "wf[time_of_flight.NumberOfSimulatedNeutrons] = 200_000  # Increase this number for more reliable results\n",
```
jl-wynen, line 63: Can you add to the docs (here or in data.py) the number used for the table? Or is it encoded in the hdf5 file?

nvaytet: It isn't. We should probably make it a coordinate, like the other metadata?

jl-wynen: Yes, but it should be unaligned.

jl-wynen: True. But it can still be good to encode this kind of metadata in the file. But that is a longer discussion as we know...

For now, just add it to the docs so we can reproduce it later.

PR #209, opened 2025-10-24
fix: convert unlimited TOA to expected range

jokasimr:
Fixes one aspect of https://github.com/scipp/essdiffraction/issues/161, specifically it converts the monitor "tof" values to the `event_time_offset` range that we expect them to have.

### DIFF
--- a/tests/dream/io/geant4_test.py
+++ b/tests/dream/io/geant4_test.py
@@ -10,8 +10,19 @@ import scipp as sc
 import scipp.testing
 import scippnexus as snx
 
-from ess.dream import DreamGeant4ProtonChargeWorkflow, data, load_geant4_csv
-from ess.powder.types import Filename, NeXusComponent, NeXusDetectorName, SampleRun
+from ess.dream import DreamGeant4ProtonChargeWorkflow, data, io, load_geant4_csv
+from ess.powder.types import (
+    CaveMonitorPosition,
+    Filename,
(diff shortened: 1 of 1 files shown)


jl-wynen (commented):
What is the range in the file?

jokasimr:
> What is the range in the file?

The range in the file is 0 to 260 ms, so ~4 pulse periods, but in most of that range the monitor intensity is zero.

jokasimr:
> I ran the user guide notebooks and most of it looks basically unchanged. But the I(d) plot in the basic user guide is bad now:

The cause is that ` normalize_by_monitor_histogram` does not take into account the case when the monitor histogram is 0. This causes division by zero for some wavelength values, and that removes the intensity from all dspacing bins that has contributions from that wavelength value.

The exact reason why there exist some zero intensity bins is the `MonitorTofData` after this change, but not before, is unclear to me. But I think it has to do with the details of how the `TofLookupTable` is applied to histogrammed data. See https://github.com/scipp/essreduce/blob/5391cd2c264ffec7cbb0990cdef526266e762c04/src/ess/reduce/time_of_flight/eto_to_tof.py#L94.

jokasimr:
result

Solid blue is old, and transparent blue is new.

There are some differences in the high-dspacing region, but that is expected since before we were dropping monitor intensity for the longest wavelengths.

PR #215, opened 2025-11-19
feat: add lorentz_correction to powder workflow

jokasimr:
Adding the correction to the workflow graph was more messy than I expected.

## Workflow/graph related issues

[The guidelines](https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#c-6-domain-types-for-data-products) say that Lorentz correction should be applied to `CorrectedDetector`.

But at that step in the workflow the events don't have `dspacing` (and Lorentz correction is a function of `dspacing`).
I don't know if this is a bug in EssDiffraction or if it is a bug in the guidelines.
I worked around the issue by removing `CorrectedDetector`, going directly from `WavelengthDetector` to `DspacingDetector`.

However, this makes the provider `add_coords_masks_and_apply_corrections` do a lot of things at once.
And it takes a lot of arguments that are not really related, so it's not very useful for understanding what the provider does.

The other option would be to add more domain types where each step is done separately. But that would add a lot of unnecessary domain types to the graph.

There's a tension between making the graphs show relevant intermediate results, and making the graphs explain what is being done to the data.

## Issues in `apply_lorentz_correction`

The existing `apply_lorentz_correction` function could not be applied to data with variances.
For now I just replaced it with something that works with variances.

### DIFF
--- a/src/ess/powder/conversion.py
+++ b/src/ess/powder/conversion.py
@@ -229,7 +229,7 @@ def add_scattering_coordinates_from_positions(
 
 
 def convert_to_dspacing(
-    data: WavelengthDetector[RunType],
+    data: sc.DataArray,
     graph: ElasticCoordTransformGraph[RunType],
     calibration: CalibrationData,
 ) -> sc.DataArray:


jokasimr:
> I there a reason why we can't compute a dspacing coord earlier and assign it to WavelengthDetector? Then we could merge WavelengthDetector and DspacingDetector and apply the corrections afterwards. (Note that we are not binning in any of these steps anyway.)

Yeah that's also a possibility. It has the same drawbacks as the other version, and it makes the name `WavelengthDetector` is a bit strange. But I agree it's a nicer split overall!

Another option is to also remove `WavelengthDetector` and do everything in one step:
Compute the coords, add the masks, and do the lorentz correction.

src/ess/powder/conversion.py
```diff
@@ -223,30 +224,31 @@ def add_scattering_coordinates_from_positions(
...
     return WavelengthDetector[RunType](out)
 
 
 def convert_to_dspacing(
-    data: CorrectedDetector[RunType],
+    data: WavelengthDetector[RunType],
     graph: ElasticCoordTransformGraph[RunType],
     calibration: CalibrationData,
-) -> DspacingDetector[RunType]:
+) -> sc.DataArray | sc.Dataset:
```
jl-wynen: How can this be a dataset?

jokasimr: I don't know. The type checker told me it could be.

Issue #217, opened 2025-12-01
BEER workflow and docs fixes


jokasimr:
There are a number of improvements we should make to the Beer workflows:

* `tc` should be better documented in the place where it is set in the loader. The name should probably change altogether to make it more clear what it represents.
* We might be able to replace the `chopper_delay_from_mode` providers by using the `lambda` parameter from the McStas simulations.
* The difference between `L0` and `Ltotal` should be better documented.
* The loaded data should have a `position` coord, to make the 3D detector view work.
* We should bin the data by pixel after loading to make the processing of the simulated data more similar to the processing of data from real files.

PR #224, opened 2026-01-16
Bin by pixel

jokasimr:
Addresses part of #217 by binning the McStas data by "pixel" after loading and adapting the workflow to handle the pixel-binned data structure.


jl-wynen (commented):
I don not see any pixel binning here, only binning by position. Is the detector aligned with the instrument coord system? And does the mcstas file not contain pixel ids?

PR #231, opened 2026-01-30
Produce better reduced data in nightly runs

nvaytet:
In the nightly runs, we upload an artifact which contains a reduced CIF file from the Dream Geant4 simulation.
The data we used to save was:
- from the `endcap_forward` detector (because the test was using a parametrized fixture and the last file that was written was from the endcap, overwriting files from previous banks)
- using the 'small' data files used in unit tests.

As a result, there has very little or no signal, and looked nothing like a d-spacing spectrum:
Screenshot_20260130_093625

Here, we use the data files with more events, and use the `mantle` detector bank.
We also increase the number of bins from 200 to 2000, to get something that analysis hopefully can make better use of.
Screenshot_20260130_093719

This was discovered by @AndrewSazonov

### DIFF
--- a/docs/user-guide/beer/beer_modulation_mcstas.ipynb
+++ b/docs/user-guide/beer/beer_modulation_mcstas.ipynb
@@ -15,6 +15,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "%matplotlib ipympl\n",
+    "import plopp as pp\n",
     "import scippneutron as scn\n",
     "\n",
     "from ess.beer import BeerModMcStasWorkflow, BeerModMcStasWorkflowKnownPeaks\n",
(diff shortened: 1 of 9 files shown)


tests/dream/geant4_reduction_test.py
```diff
@@ -304,9 +288,26 @@ def test_pipeline_can_save_data(workflow):
     _assert_contains_tof_data(content)
 
 
-def test_pipeline_save_data_to_disk(workflow, output_folder: Path):
-    workflow = powder.with_pixel_mask_filenames(workflow, [])
```
SimonHeybrock: Is this the fixture you mentioned? If it leads to some unintentional test coupling/interaction it is worth investigating that better. What is the fixture scope? Should workflows from the fixture get copied instead of modified directly?

PR #262, opened 2026-03-25
Beer powdiff workflow

### LINKED ISSUE #235: [Requirement] BEER: Powder diffraction reduction workflow
### Executive summary

Implement reduction worklow for Powder Diffraction at BEER

The enclosed screenshot describes the different steps

Image

For this first implementation, we use 2D detectors

Follow-up: support of 3D detectors

### Inputs

McStas NeXus files for sample in container, empty container and Vanadium

### Methodology

See actions above and existing DREAM workflow

### Outputs

**Data**
2 1D spectra in ToF: a spectrum per detector bank 
Format: xye

**Reduction script**
The notebook should be available in Code Shelf

Integrated into reduction workflow

### Test cases

Please contact me to have access to the 3 NeXus files

### Existing implementations

Many steps done for DREAM can to be applied to BEER data
Please contact me to have access to the draft of the workflow

### Comments

The resolution of the detectors has not been decided yet

jokasimr:
Fixes #235


src/ess/beer/conversions.py
```diff
@@ -185,17 +189,48 @@ def _tof_from_dhkl(
     return out
 
 
+def t0_estimate(
+    wavelength_estimate: sc.Variable,
+    L0: sc.Variable,
+    Ltotal: sc.Variable,
+) -> sc.Variable:
+    """Estimates the time-at-chopper by assuming the wavelength."""
```
SimonHeybrock: - Not clear to me what L0 is without context. Can you explain in the docstrings? Either here or module-level?
- Could a helper from `scippneutron` be used for this conversion?

src/ess/beer/conversions.py
```diff
@@ -185,17 +189,48 @@ def _tof_from_dhkl(
...
+
+def tof_from_t0_estimate_graph(
+    da: RawDetector[RunType],
+    gg: GeometryCoordTransformGraph,
+) -> ElasticCoordTransformGraph[RunType]:
+    """Graph for computing ``wavelength`` in pulse shaping chopper modes."""
+    return {
+        **gg,
+        't0': t0_estimate,
+        'tof': lambda time_of_arrival, t0: time_of_arrival - t0,
```
SimonHeybrock, line 222: Can you clarify why we still have `tof` here despite the switch to use the new `wavelength`-LUT in other code above?

jokasimr, line 222: Eventually we will replace this mechanism for computing wavelengths in beer pulse shaping modes with the lookup table approach to make it consistent with the other techniques.

src/ess/beer/io.py
```diff
@@ -166,12 +176,23 @@ def _load_beer_mcstas(f, north_or_south=None, *, number):
         positions['MCB'],
         positions['MCC'],
     )
+    data = (
+        next(_find_all_h5(data_dir, f'.*{north_or_south}{number}'))
+        if north_or_south is not None and number is not None
+        else next(_find_all_h5(data_dir, f'.*{north_or_south}'))
+        if north_or_south is not None
+        else next(_find_all_h5(data_dir, f'/entry1.*bank.*{number}'))
```
jokasimr, line 184: The McStas file loading is complex because there have been several rounds of changes to the McStas output files, and we want to be "backwards compatible" with the old McStas output structure because otherwise we have to replace the files in the tests/docs.

I've raised this with Celine and I don't expect there will be more significant changes going forward.

SimonHeybrock (commented):
Small questions:

src/ess/beer/conversions.py
jokasimr: Beer currently does not use the LUT approach to compute wavelengths. Eventually we will move to that approach for the pulse shaping modes (the modulation modes are incompatible with the LUT approach).

SimonHeybrock: But I saw a lot of changes that replaced Tof with Wavelength?

jokasimr: Yes, the workflow is still using the generic unwrap workflow which does not have any public `Tof...` types.

SimonHeybrock: So where does TOF come into play? Is it TOA->TOF->Wavelength, or TOA->Wavelength->TOF as for DREAM final result, or something else?

jokasimr: It's toa->tof->wavelength.

SimonHeybrock: So what is `GenericUnwrapWorkflow` being used for? As far as I understand it does not support TOF anymore.

jokasimr: It's used so that the interface of the workflow matches what it will look like once we have lookup tables for the pulse skipping modes. That the wavelengths are currently not computed using lookup tables is an implementation detail, it does not need to be reflected in the public interface.

src/ess/beer/conversions.py
jokasimr: I don't think it makes it easier to understand or more correct to mix in a helper from scippneutron here.

This will anyway be replaced by the lookup table approach eventually.

I replaced `L0` by `moderator_to_detector_distance`.

SimonHeybrock: 👍 better name... but now unclear how that differs from `Ltotal`?

jokasimr: `Ltotal` is here the distance from the chopper that determines wavelength, the distance from the "effective source".

SimonHeybrock: Where is that written down? Is that the same definition that is used at instruments that use the `essreduce.unwrap` module?

jokasimr: In the unwrap module `Ltotal` is the distance from the detector to the "source position" of the lookup table.
The "source position" of the lookup table is the position where the time-wavelength distribution is equal to the source distribution.
For the source distributions we use I think that position is typically the beam port of the instrument, but that is worth double checking with Neil.

It does not really matter because when we load non-mcstas files then `Ltotal` will be determined by the generic workflow, it's only in the McStas workflow where it is defined manually.

jokasimr: I made the wavelength defining chopper position a separate quantity to make it more clear what the function does.

This function could in principle be used with non-mcstas data, as long as `wavelength_estimate` and the chopper position is set on the data.

SimonHeybrock (approved):

